PR #15, opened 2020-05-12
Allowing word definitions without a part-of-speech

yolpsoftware:
There are Wiktionary pages where the word definitions (glosses) are not organized inside Part-Of-Speech sections. See for example the Chinese section of

https://en.wiktionary.org/wiki/%E6%83%B3#Chinese

Although 想 is used exclusively as a verb, there is no subsection with a "Verb" title. This results in Wiktextract reading no `sense` entries from this page.

To circumvent this, I changed the code to allow no-POS headers, currently only for Chinese, and only for sections with title "Definitions".

Line 2409 should probably be extended to allow this for more languages (maybe all?).

### DIFF
--- a/wiktextract/page.py
+++ b/wiktextract/page.py
@@ -2406,11 +2406,12 @@ def page_iter(config, text):
                 # a new language or a misspelling or a previously unsupported
                 # subtitle.
                 sectitle = sectitle.lower()
-                if sectitle in part_of_speech_map:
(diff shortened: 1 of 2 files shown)


tatuylonen:
Recognition of "Definitions" section is has been implemented in the current version for some time.  Thanks for the report though (I wish I'd followed the pull requests more closely).

PR #17, opened 2020-05-12
Bug: examples are duplicated over all senses

yolpsoftware:
This can be demonstrated well with the Italian word [trovare](https://en.wiktionary.org/wiki/trovare#Italian):

image

Even though the example sentence _"lo chiamo spesso, ma mai lo trovo"_ is valid only for the third sense, Wiktextract includes this example sentence in all senses. This is clearly wrong, the example does not make sense when the word is translated as "to find".

### DIFF
--- a/wiktextract/page.py
+++ b/wiktextract/page.py
@@ -1680,13 +1680,13 @@ def parse_preamble(config, data, pos_sectitle, text, p):
 
     # Parse word senses for the part-of-speech.
     for node in p.lists():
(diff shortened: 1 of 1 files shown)


yuchenlin:
This is nice!

meetDeveloper:
Has this been fixed @tatuylonen. I am unable to find duplicate examples for the word trovare, seems like it has been fixed.

PR #18, opened 2020-05-13
Saving Chinese characters in Unicode / UTF8

yolpsoftware:
Currently, the default value of the parameter `ensure_ascii` of `json.dumps` leads to all non-European characters being escaped. Instead of

```
"word": "馬",
```

we get

```
"word": "\u99ac",
```

This PR changes that, so that we get nice Unicode / UTF8 characters in the output.

### DIFF
--- a/wiktwords
+++ b/wiktwords
@@ -159,7 +159,7 @@ if __name__ == "__main__":
             out_tmp_path = out_path
         else:
             out_tmp_path = out_path + ".tmp"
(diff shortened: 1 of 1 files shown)


tatuylonen:
I reimplemented this change for the current version.  I think it is a good change and makes the output much more readable.  (The cost is that people using non-unicode environment need to explicitly specify encoding="utf-8" in open, which may cause some confusion).  However, I think the readability is worth it.

PR #24, opened 2020-07-30
Feature/hierarchical glosses

erolm-a:
This pull request adds the following:

1. Hierarchical senses. The `senses` key keeps a similar structure to before, but also adds subsenses for each item if available; they follow the same structure as senses, and usually start with `##` (nested numbered bullet point). Each subsense also can have `usages`, which in turn follow the same structure of senses and subsenses. As a side effect, tags are not squashed together but hierarchically separated as well.

2. Basic quotation support. Because some of the nested bullet points are not senses, there must be some disambiguation logic to prevent examples or quotations to be counted as glosses.

3. Better example extraction. Some examples do not use the ux template, but as documented [here](https://en.wiktionary.org/wiki/Wiktionary:Example_sentences) this is not a hard requirement.

This PR is still a WIP project and needs further review and testing, but results are promising.

By and large, the latest changes add roughly 15-20 MB of information into the JSON output.

- first-level examples: 93735 (of which 47298 non-empty)
- second-level examples: 2433 (of which 621 non-empty)
- third-level examples: 64 (of which 4 non-empty)

CREDITS:

This PR leverages some of the work carried out by @yolpsoftware in a number of his PRs.


tatuylonen:
Implemented differently in current architecture.

PR #50, opened 2021-05-14
Fix #49

### LINKED ISSUE #49: Head type la-num-adj missing
The head type `la-num-adj` is not exported, seen for example on this page: https://en.wiktionary.org/wiki/unus#Latin (the numeral entry).

I believe this is because the `head_tag_re` regex misses `num-adj`.

fpw:
Tested and confirmed that this fixed #49


fpw:
Since la-pronoun is also missing, I'll start a new PR.

PR #98, opened 2021-11-18
Update README.md

rgossiaux:
Fix --cache-file => --cache in Readme

The --cache-file option gives me an error and doesn't seem to exist.


tatuylonen:
Thanks

PR #143, opened 2022-08-04
Ignore Spanish Wiktionary Wikcionario pages

xxyzz:
I'm trying to parse the Spanish dump file and find the command takes a long time parsing [Wikcionario:Frecuencias/Finés](https://es.wiktionary.org/wiki/Wikcionario:Frecuencias/Fin%C3%A9s/000001-010000) pages. These pages are not word pages and this pull request ignores them so the command won't waste over ten hours parsing them.

### DIFF
--- a/wiktextract/wiktionary.py
+++ b/wiktextract/wiktionary.py
@@ -54,8 +54,8 @@ def page_handler(ctx, model, title, text, capture_cb, config_kwargs,
     # steps.
     title = re.sub(r"[\s\000-\037]+", " ", title)
     title = title.strip()
-    if capture_cb is not None:
-        capture_cb(model, title, text)
+    if capture_cb and not capture_cb(model, title, text):
+        return None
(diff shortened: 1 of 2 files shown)


kristian-clausal (commented):
You are absolutely correct in that capture_cb is a bit weird here. At a quick glance, capture_cb returns capture_page(...), which returns a boolean about whether a page should be analyzed normally (True) or specially (False), but nothing seems to be done with that return value. I'll ask Tatu about whether there's something going on here, or whether it's just a bit of old code that needs to be tidied up.

kristian-clausal:
Tatu checked it out, and he agreed. We've been silently analyzing pages that should not have been, so fingers crossed there will be less bad data now, so thanks for catching this! We didn't see any problem with adding "Wikcionario:" into the ignored prefixes, and we also removed "Reconstruction" from ignored prefixes from our side while at it.

Issue #133, opened 2022-04-26
Please use .jsonl suffix


1over137:
The format you are using (which is a perfectly good choice) is commonly known as `jsonlines` (https://jsonlines.org/)
Since it is technically not a json file, it would be most appropriate to use the .jsonl suffix instead, so that people and programs can deal with it properly.

alexchandel:
Using `.jsonl` on the kaikki.org downloads would be helpful.  I didn't see anywhere in the codebase to change this, so it must be somewhere in Tatu's CI's.

kristian-clausal:
See upcoming change issue #603. Rejoice, .jsonl enthusiasts!

PR #149, opened 2022-08-08
Explicitly note JSON Lines output format

alexchandel:
Related: #133

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ to view the JSON produced for each page.  You can download all data,
 data for a specific language, data just a single word, or data for a
 list of related words (e.g., a particular part-of-speech or words
 relating to a particular topic or having a particular inflectional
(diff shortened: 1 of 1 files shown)


alexchandel:
Is this good to merge?

kristian-clausal:
At first I thought we were breaking some kind of requirement by not naming the files with .jsonl, but turns out that's just a "suggestion"; the three requirements are utf-8

- [x] utf-8
- [x] each line is valid JSON
- [x] line separator is '\n'

I don't see why not this small clarification couldn't be made.

1over137:
I suggest you reconsider this. While it doesn't break a requirement to name it as .jsonl, it does break the expectation that a file named x.json is a valid json file. These are not valid json files, and will produce an error in any parser. Using .jsonl hints users and programs that this is something other than a json file.

PR #152, opened 2022-08-11
avoid crashes with illegal --num-threads option

### LINKED ISSUE #151: crash with --num-threads set
This may be another OSX thing, but I've found that I cannot use the `--num-threads` option, as it always leads to a crash.

Example run command:

    ./wiktwords --all --language Vietnamese --pages-dir pages --num-threads 2 --cache /tmp/wikt-cache --out vn_out.jsonl enwiktionary-latest-pages-articles.xml.bz2 | tee log_vn_1.txt

And here's the error output:

    multiprocessing.pool.RemoteTraceback:
    """
    Traceback (most recent call last):
    File "/usr/local/Cellar/python@3.9/3.9.10/Frameworks/Python.framework/Versions/3.9/lib/python3.9/multiprocessing/pool.py", line 125, in worker
        result = (True, func(*args, **kwds))
    File "/Users/nate/dev/workspaces/python_workspace/wiktextract/env/lib/python3.9/site-packages/wikitextprocessor/core.py", line 60, in phase2_page_handler
        ctx.start_page(title)
    AttributeError: 'NoneType' object has no attribute 'start_page'
    """

    The above exception was the direct cause of the following exception:

    Traceback (most recent call last):
    File "/Users/nate/dev/workspaces/python_workspace/wiktextract/./wiktwords", line 277, in <module>
        parse_wiktionary(ctx, args.path, config, word_cb, capture_cb,
    File "/Users/nate/dev/workspaces/python_workspace/wiktextract/wiktextract/wiktionary.py", line 134, in parse_wiktionary
        return reprocess_wiktionary(ctx, config, word_cb, capture_cb)
    File "/Users/nate/dev/workspaces/python_workspace/wiktextract/wiktextract/wiktionary.py", line 148, in reprocess_wiktionary
        thesaurus_data = extract_thesaurus_data(ctx, config)
    File "/Users/nate/dev/workspaces/python_workspace/wiktextract/wiktextract/thesaurus.py", line 254, in extract_thesaurus_data
        for word, linkages in ctx.reprocess(page_handler, autoload=False):
    File "/Users/nate/dev/workspaces/python_workspace/wiktextract/env/lib/python3.9/site-packages/wikitextprocessor/core.py", line 1389, in reprocess
        for success, title, t, ret in \
    File "/usr/local/Cellar/python@3.9/3.9.10/Frameworks/Python.framework/Versions/3.9/lib/python3.9/multiprocessing/pool.py", line 870, in next
        raise value
    AttributeError: 'NoneType' object has no attribute 'start_page'

garfieldnate:
Exit with an error message if the user tries to set `--num-threads`
on a system where it leads to program errors. Killing the program
early with a helpful message is much preferable to the crash with
stack traces emitted several minutes into processing (at the
beginning of phase 2).

Fixes #151.

### DIFF
--- a/wiktwords
+++ b/wiktwords
@@ -197,6 +197,12 @@ if __name__ == "__main__":
     else:
         print("Capturing words for:", ", ".join(args.language))
 
+    if args.num_threads and args.num_threads > 1:
+        import multiprocessing
+        if multiprocessing.get_start_method() == "spawn":
+            print("--num-threads not supported on this OS (no stable implementation of fork() available)")
+            sys.exit(1)
+
     # Open output file.
     out_path = args.out
     if out_path and out_path != "-":


kristian-clausal (approved):
This seems correct, and importing multiprocessing here or later in core.py probably shouldn't affect performance that much (if at all, considering it should just use the cached import).

Issue #153, opened 2022-08-14
Citation templates being included in etymology_templates lists


jmviz:
In `page.py`, there is effort to avoid including citation templates in `etymology_templates` lists (see [`ignored_etymology_templates_re`](https://github.com/tatuylonen/wiktextract/blob/32bd4ffec21552efae3cd8ccda9a6f2e73782a56/wiktextract/page.py#L518) and its use in [`etym_post_template_fn()`](https://github.com/tatuylonen/wiktextract/blob/32bd4ffec21552efae3cd8ccda9a6f2e73782a56/wiktextract/page.py#L2531). However, I noticed there are quite a few citation templates being included in the output. Here is a non-exhaustive list of some I noticed:
```
archivedotorg
cite book
Cite book
cite journal
cite web
ISSN
lang
num
R2A
w
```
Since these templates generally are not in the wikitext itself, I understand they are being added by the recursive expansion of templates invoked by templates that are in the wikitext. Often though, the only citation templates that are in the wikitext are ones that are currently ignored, like `R:*` templates. So, while any `R:*` template won't itself be added to `etymology_templates`, all the templates invoked by it (that don't match `ignored_etymology_templates_re`) will be, which seems strange. 

I would think there must be a way to disallow such behavior by tweaking the `template_fn`s and `post_template_fn`s passed to `clean_node()`, but I wasn't able to figure it out in the brief time I spent on it.

Of course, the easiest fix would be just to add the templates I listed to the ignored list. But there are surely others I missed that don't belong as well.

kristian-clausal:
I've committed a temporary kludge to this issue, based on what Tatu recommended, but I'm unhappy with it. At least this should fix stuff like in [your mileage may vary](https://en.wiktionary.org/wiki/your_mileage_may_vary), where the cite-journal template was broken up into subtemplates like ISSN, as you've noted, but I am pretty sure the *intent* of the code has been (as far as I understood) for non-etymology templates being fed into a normal header expansion function that does something 'normal' with it, ie. the citation should appear somewhere. But at least now we don't have garbage data in the etymologies.

[...]

PR #155, opened 2022-08-15
deal with cite template redirects

jmviz:
Change ignored_etymology_templates to match more cite template names.


kristian-clausal:
I've made a different commit that doesn't change the first bit of the compiled regex, but adds a bunch of these redirects into the ignored_etymology_templates manually. See #153 for more.

PR #165, opened 2022-10-05
clarify lang_code as headword language

garfieldnate:
The old wording made me think that `lang_code` was for the language of the glosses (the language of the Wiktionary edition where the entry is found).


tatuylonen:
This has now been merged. It is a good clarification.

PR #179, opened 2022-11-09
Fix parse translation subpage error and add form of tags for Chinese Wiktionary

xxyzz:
The first commit only fix the error while parsing the "see translation subpage" template, I only remember the code failed at a `assert` but forget which line it failed. There are many `more than one value in "roman"` debug logs but I'm not dare to touch the code in [`parse_translation_desc()`](https://github.com/tatuylonen/wiktextract/blob/4ff24e2fa2e152a2ffeb5058f65cb12270207eeb/wiktextract/form_descriptions.py#L1854).

The second commit adds some "form of" tags if the page uses [those templates](https://en.wiktionary.org/wiki/Category:Form-of_templates). IIUC, the current code adds those tags by checking the expanded template text, which is only works for the English dump file and complicated.


kristian-clausal (approved):

kristian-clausal:
If you are making commits that do not belong logically with others in a pull request, like the template_nesting fix here, please make them their own pull request for the master branch and not part of sidebranches that need further verification.

xxyzz:
I happen to run the same branch on a server which has run time limitation and I forget to checkout a new branch because I have to fix them quickly before the server shut down. I will separate commits in future pull requests.

Issue #196, opened 2022-12-28
Replace "nose" with a newer test framework


yacoder:
"nose" seems broken since Python 3.10 changed something in collections: https://github.com/nose-devs/nose/issues/1122
And generally "nose" is not maintained.

I'm not an expert in Python test frameworks so I won't suggest a replacement, but maybe I can help migrate if you have a preference for a replacement framework.

kristian-clausal:
We're definitely using nose2 for testing here at Clausal, but requirements.txt does have just "nose" in it. I'm not really familiar with the pitfalls of this, so I'll ask Tatu about it after New Years. The Makefile has nose2. Changes probably just involve a change in requirements.txt and documentation, but better safe than sorry.

kristian-clausal:
Fixed in #197

PR #197, opened 2022-12-28
Add Windows VS Code instructions

yacoder:
I was able to run tests via VS Code on Windows following these instructions. 

There is only one error, related to bz2 decompression, no `/bin/sh`, but that's a known problem on Windows, I'll try to fix it next, so that all tests can pass.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ pages
 .coverage*
 kj-wiktwords.sh
 tmp/
+.venv/
+.vscode/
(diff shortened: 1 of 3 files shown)


kristian-clausal:
This seems all good, but I'll wait until Tatu has had a chance to look at explicitly changing over to nose2. Thanks for the pull request! It is very helpful to have someone look at this from the Windows side.

yacoder:
Ok, thanks :) I'm planning to send more, as I'm trying to use it to extract data from some Slavic languages wiktionaries.

yacoder:
Also switched to nose2 and updated the doc, works on my machine :)

Should fix #196

kristian-clausal:
Tatu ok'd switching explicitly over to nose2 (it was just an oversight), this all seems ok to me.

PR #201, opened 2023-01-18
Add languages.json for Czech and Slovak

Vuizur:
This commit adds code to generate languages.json for Czech and Slovak. I haven't run Wiktextract on these languages yet, but maybe the code is already useful to others who want to work on other Wiktionaries because it shows how to parse the source code instead of expanding the templates (because sometimes you can't expand them as there is no function to invoke).


kristian-clausal:
Looks good, I don't see why not to add it right away.

Vuizur:
Thanks!

PR #203, opened 2023-01-21
Loading the dump into a SQLite DB

Vuizur:
I think for a lot of applications it is useful to have the data as an SQLite file, mostly because it allows searching for words in 1 - 2 milliseconds and to filter things like languages. This script generates the columns automatically so there shouldn't be a lot of maintenance burden. It also demonstrates how to get a proper collation because by default SQLite is pretty weak there and doesn't even support case insensitive search for cyrillic, for example. (And because I found it to be a somewhat confusing topic.)

Additionally I added a way to not only search the words but also the inflections for words (both diacritic/case insensitive). It seems to work with the entire dump. From the 15 GB dump you get a 23 GB database (due to indexing).

### DIFF
--- a/usertools/sqlite_dictionary.py
+++ b/usertools/sqlite_dictionary.py
@@ -2,6 +2,8 @@ import json
 import os
 import sqlite3
 import unicodedata
+import tqdm
+import logging
 
 
 def strip_accents(s: str):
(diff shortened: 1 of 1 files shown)


kristian-clausal:
Now that @xxyzz has committed code that migrates us from wikt-cache to Sqlite, would it be possible for you to check out if it has any of the issues you've come across when working on your branch, or to integrate some of the functionality you have?

Vuizur:
I checked the code again, I think it is ready now. In my opinion SQLite is in general a good option - lots of people know it and it's fast. But having a SPARQL example with oxigraph with some useful queries would also be really cool and educative.

I also once experimented with loading the JSON into MongoDB, this is also a decent approach, because it allows super easy index creation for the nested objects of the JSON.

xxyzz:
My Proficiency repository has some SPARQL queries: https://github.com/xxyzz/Proficiency/blob/aa9260a868e1846e81e8c30e33fcc6a4a970c3f7/extract_dbnary.py#L81

usertools/sqlite_dictionary.py
```diff
@@ -0,0 +1,162 @@
...
+    if s1 < s2:
+        return -1
+    elif s1 > s2:
+        return 1
+    else:
+        return 0
+
+
+class WiktionaryDictionary:
+    """This class is a wrapper around a sqlite database containing the Wiktionary dump.
```
kristian-clausal, line 57: Because of the use of the word "dump", I was confused that this whole thing was about creating a database out of the Wiktionary dump file... That is, the big zip files we get from Wiktionary that contains the source text files. The extracted data should probably be called something else to avoid confusion. `Wiktextract data` or something similar should work.

PR #229, opened 2023-04-12
process Descendants sections

### LINKED ISSUE #148: [error] derived terms are skipped in hierarchical entries (like Proto-Indo-European)
Unlike living languages, Proto-Indo-European words' derived terms are given in a hierarchical list, rather than a flat list.  But wiktextract/kaikki only picks up the "deepest" entry.  For example, in the PIE term [sek-](https://en.wiktionary.org/wiki/Reconstruction:Proto-Indo-European/sek-), the first derived term (reproduced below) should be "sek-eh₂-yé-ti" and the second "sekajō", but these is skipped for "secō".  These intermediate derived terms should not be skipped over.

* sek-eh₂-yé-ti or *sek-h₁-yé-ti
  * Proto-Italic: *sekajō
    * Latin: secō (_see there for further descendants_)

jmviz:
This PR serves as a proposal implementation for the processing of Descendants sections. It would close #148; see there for an example of what's currently lacking and what this code produces. I would be open to reworking any aspect of the data format or code based on feedback.

### Motivation
I have been using this for local processing for some time. My use case is the creation of etymological trees/graphs using wiktextract data. Using solely wiktextract's `etymology_(templates|text)` data to infer etymological relationships between items can be unreliable, due to the relative lack of structure in Etymology sections' wikitext. In Descendants sections, especially the ones that take the form of hierarchical lists, inferring etymological links between words is more reliable, as the etymological tree structure is directly provided in the wikitext. Thus, using this data as a complement to the Etymology data proves highly valuable. In general, anyone who is using wiktextract's `etymology_(templates|text)` data would likely stand to gain from having access to Descendants section data as well, in terms of the accuracy and completeness of whatever secondary data they are generating.  

### Data format
The code produces a new top-level key in a wiktextract item's JSON, `descendants`, which is an array of objects corresponding to individual lines in the list of descendants. Each descendant object contains the following keys:

* `depth`: The level of indentation of the current line. This can be used to track the hierarchical structure of the list. 
* `templates`: An array of objects corresponding to templates that appear on the line. The structure of each of these objects is the same as the structure of each object in wiktextract's `etymology_templates`.
* `text`: The expanded and cleaned line text, akin to wiktextract's `etymology_text`. 

Obviously, I modeled the data for each line after wiktextract's etymology section data. This works well enough, although for the simple Descendants sections that are just flat lists of single `{{desc}}` templates in can be a bit repetitious. Also, `{{desctree}}` is a somewhat common template in these sections, and its expansion generates an HTML hierarchical `<ul>`, the structure of which gets lost in this expansion as the HTML is cleaned out. I didn't address this, as it seemed leaving in arbitrary HTML would be antithetical to the general wiktextract data format, and there wasn't an immediately obvious plain-text alternative. The HTML-less expanded text might still be useful for some applications. 

### Where it is called
This produces a `descendants` object for `Descendants` sections of any items, and also on `Derived terms` and `Extensions` sections of items in `Reconstruction` languages that have PoS `Root`, as these sections have the same format. For this latter special case, if multiple of these sections appear, all of the data is combined into one `descendants` object.


kristian-clausal:
Seems ok to me. I'm going to merge it and we'll see if it breaks kaikki by tomorrow.

Doesn't have to be perfect, and this is a good start. I see what you're doing with the depth, but it feels a bit kludgey. If the list is out of order (JSON list order should be preserved, but still), or you take one item out of the list so that it doesn't have context, you can't access its ancestors. On the other hand, preserving the data by duplicating parent data (in a list) seems like needless duplication. Nesting everything could work.

PR #230, opened 2023-04-12
don't classify desc as romanization if it is in data["categories"]

jmviz:
Noticed some spurious romanizations. Consider for example [Reconstruction:Latin/sufferio](https://en.wiktionary.org/wiki/Reconstruction:Latin/sufferio).  `parse_word_head()` was going through the cleaned version of `{{la-verb|4|*sufferiō|*sufferīv|*suffert}} {{lb|la|Proto-Romance}}`, which is `*sufferiō (present infinitive *sufferīre, perfect active *sufferīvī, supine *suffertum); fourth conjugation (Proto-Romance)`, and was determining `Proto-Romance` to be a romanization. This can be ruled out as a romanization since `{{lb}}` generates the category `Proto-Romance`, which is in `data["categories"]`. So this PR adds a check to make sure cases like this are avoided. 

Now, the spurious romanization is gone, but there now is a `DEBUG: unrecognized head form: Proto-Romance` that gets logged as the `desc` ends up not being classified. I left this alone as I'm not sure what the general system is for these classifications. (It's possible this check should be moved up higher in the `for desc_i, desc in enumerate(new_desc):` loop, depending what that system is...)

(There happens to be an unrelated Lua execution error on this page when invoking `{{VL-conj-4th}}`, but that is neither here nor there for this PR.)

### DIFF
--- a/wiktextract/form_descriptions.py
+++ b/wiktextract/form_descriptions.py
@@ -1631,7 +1631,8 @@ def parse_word_head(ctx, pos, text, data, is_reconstruction,
                        continue
                     elif (cls in ("romanization", "english") and
                         not have_romanization and
-                        classify_desc(titleword) == "other"):
+                        classify_desc(titleword) == "other" and
+                        not ("categories" in data and desc in data["categories"])):
                         # Assume it to be a romanization
                         add_related(ctx, data, ["romanization"], [desc],
                                     text, True, is_reconstruction, head_group,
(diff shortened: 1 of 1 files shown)


kristian-clausal:
This seems like a sensible change; a situation where a romanization and a category are identical and present on the same exact page (especially when a vast majority of categories are at least a couple of words) should be extremely rare or even non-existent.

kristian-clausal:
After changing things to skip the debug message (by inserting the category check inside the romanization branch which would have `continue`'d the loop and bypassed the debug message), I actually realized the correct way to fix this (particular example) is to add Proto-Romance in uppercase_tags in tags.py, so that `Proto-Romance` gets picked up as a tag. That way, it doesn't end up in romanizations and stops the debug message from triggering.

But that doesn't mean the code here is wrong in general; it is still true that if we see a romanization that looks like a category from that page it shouldn't be used as a romanization. In that case, the debug message is useful information.

kristian-clausal:
Also was able to figure out the Lua thing:
https://github.com/tatuylonen/wikitextprocessor/commit/05430617d5631618cb21a1522c4d5a18d20add58
https://github.com/tatuylonen/wikitextprocessor/issues/35

It was an issue with super-weird code in Module:VL-verb; the `__tostring` metamethod was used to pass data in the form of tables (instead of strings), which is a thing that was not prohibited by 5.1, but has been prohibited in later versions of tostring() (the function calling __tostring).

PR #235, opened 2023-04-19
Adding new Wiktextract-specific context

kristian-clausal:
Because some functions in Wiktextract have much too long parameter lists, I've wanted for a while now to create more context objects to hold them, when it makes sense.

There is also a need to create a separation between Wiktextract and Wikitextprocessor's Wtp context object, called `ctx` when it's a wiktextract parameter. Wiktextract should not modify ctx, because it belongs to Wikitextprocessor, so it needs its own context. This preliminary general Wiktextract context is what's being worked on in this branch.

There is actually already something similar to a Wiktextract context in WiktionaryConfig; `config` holds some data that a context really should (like error message data stuff). However, to make things simple at first, I've created an umbrella-context object called WiktextractContext (wxr) that doesn't replace `config` but takes `config` as one of its fields. This way, there is no need actually touch WiktionaryConfig yet.

`wxr` is created (and recreated) in the same places as `config` is, copying the same methods that allow parallelization over processes.

`wxr` also has what was previously `ctx` (Wtp) as a field `wxr.wtp`. Every time ctx-dot methods and fields were previously accesses, now you do so with wxr.wtp instead.

`wxr` also already has slots for the three most common Wiktextract parameters that make sense as context fields: `word`, `lang` and `pos`.

Currently, this branch was mostly just changing ctx and config in signatures into wxr, and calls to ctx and config methods and fields were replaced by wxr.wtp and wxr.config calls. This will make it much easier for future search-and-replace changes to everything.

Testing this out with an interrupted run on Franco-Provençal seemed to indicate I got it all to work. However, if anyone wants to take a look before I merge this, go ahead. There are *so* many changes that something must have slipped through.


kristian-clausal:
I knew I forgot something; the tests need to finished!

kristian-clausal:
For the hell of it, I tried to merge this with the big changes made recently, just to see how hard it could be, and it was enough of a pain that I decided just to reimplement the changes based on current master instead. I am pretty sure this was the correct choice.

PR #239, opened 2023-04-25
Migrate  legacy`setup.py` to `pyproject.toml`

xxyzz:
This would simplify the installation process since the `wikitextprocessor` project will be installed from source automatically.


kristian-clausal:
Will be merged in one together with https://github.com/tatuylonen/wikitextprocessor/pull/36 over in wikitextprocessor, as long as https://github.com/scoder/lupa/pull/238 over in scoder/lupa gets merged so that we don't have to work with a temporary fork.

PR #245, opened 2023-05-03
get language data natively through the relevant lua module(s) in each language's wiktionary dump file

jmviz:
As far as I know, every language's wiktionary has some sort of lua module(s) like English's [Module:languages](https://en.wiktionary.org/wiki/Module:languages) which, at minimum, store all the language data for the wiktionary, and also often export functions that can be used to access specific portions of the data. Using the functionality of these modules allows us to retrieve exactly the same data as would be retrieved on wiktionary itself, and access all of the data natively and directly. This should allow for a generally more robust pipeline for language data than relying on remote API calls to invoke templates that only expose certain parts of data, or on manually parsing pages for language wiktionaries that don't even offer such templates. 

For example, for English wiktionary, [Module:list_of_languages,_csv_format](https://en.wiktionary.org/wiki/Module:list_of_languages,_csv_format) only offers `otherNames`, but this is a deprecated field; it does not offer the non-deprecated fields `aliases` and `varieties`. Many languages do not have any `otherNames` but do have `aliases` (and vice versa). But using the lua module, we can call for each language `getOtherNames()`, which returns all three of these fields collapsed into one list. As another example, if it were decided that the odd wikimedia codes do have to be dealt with somehow, they can be gotten directly by calling `getWikimediaLanguageCodes()` for each language.

So for this PR I downloaded the latest (20230501) wiktionary dumps for each language that currently has language data in `wiktextract/data/`, and wrote little lua module files that interact directly with the relevant lua modules and export the language data. I attempted to make the output as close as possible to the existing data. The main difference is that for English/Chinese, each language has as its names list the entirety of its `otherNames`, `aliases`, and `varieties`. I think `aliases` at least deserves to be included. `varieties` I was less sure about; if listing all varieties of a language as alternative names gets in the way of current special handling of varieties stuff (like in translations or wherever else), then I can remove varieties from the data. 

Other changes:

- Moved and renamed the existing `get_languages.py` script, and changed a `match` block to `if`s so that it doesn't require Python 3.10.
- Added some more checking and (optional) warning messages during the creation of `LANGUAGES_BY_NAME` to guard against / alert the user about mappings being overridden. 

N.B. This PR uses the modification to `process_dump` from https://github.com/tatuylonen/wikitextprocessor/pull/40, it shouldn't be merged unless that one is.


kristian-clausal:
We have quite a bit of other stuff going on with varieties, so those should probably not be in the main file. If that data can be separated into its own thing, that would be nice (in the future), but currently they're handled in a couple of places and a couple of lists of keywords etc. so it's a hassle.

Otherwise looks good (afaict), the changes in the language code files seem sensible, and this seems like it's a straightforward side-thing that only touches on those data files (which we can see are fine); I'll merge after you remove the variants from the code data.

PR #250, opened 2023-05-17
Fix db connection error

xxyzz:
Fixes https://github.com/tatuylonen/wikitextprocessor/issues/45.

Also fix the error and improve the speed of `extract_namespace()` which is used in the `--modules-file` and `--templates-file` option.


kristian-clausal:
Looks good :+1:

PR #254, opened 2023-05-29
Support debugging multiple pages with the `--page` command argument

xxyzz:


kristian-clausal:
Testing out, giving --page malformed input (for example, `--page pages/words/etu.txt pages/words/taka.txt` or `--page etu taka` where there is no existing file or file in the database results in a cryptic traceback dump and could do with some better error messaging.

```
Capturing words for all available languages
2023-05-29 08:56:31,099 INFO: First phase - extracting templates, macros, and pages
2023-05-29 08:56:31,099 INFO: skip_extract_dump: True, save_pages_path: None
Traceback (most recent call last):
  File "/home/kristian/Repos/wiktextract/wiktextract/wiktwords.py", line 439, in <module>
    main()
  File "/home/kristian/Repos/wiktextract/wiktextract/wiktwords.py", line 331, in main
    process_single_page(title_or_path, args, wxr, word_cb)
  File "/home/kristian/Repos/wiktextract/wiktextract/wiktwords.py", line 67, in process_single_page
    ret = parse_page(wxr, title, text)
  File "/home/kristian/Repos/wiktextract/wiktextract/page.py", line 3479, in parse_page
    tree = wxr.wtp.parse(text, pre_expand=True,
  File "/home/kristian/Repos/wikitextprocessor/wikitextprocessor/core.py", line 1737, in parse
    text = self.expand(
  File "/home/kristian/Repos/wikitextprocessor/wikitextprocessor/core.py", line 1400, in expand
    expanded = expand_recurse(encoded, parent, not pre_expand)
  File "/home/kristian/Repos/wikitextprocessor/wikitextprocessor/core.py", line 1215, in expand_recurse
    if not self.template_exists(name):
  File "/home/kristian/Repos/wikitextprocessor/wikitextprocessor/core.py", line 1664, in template_exists
    self.get_page(name, self.NAMESPACE_DATA["Template"]["id"])
  File "/home/kristian/Repos/wikitextprocessor/wikitextprocessor/core.py", line 1617, in get_page
    for result in self.db_conn.execute(
sqlite3.ProgrammingError: Cannot operate on a closed database.
```

xxyzz:
I add a error message in https://github.com/tatuylonen/wiktextract/pull/255. The `--path` argument should be used like this: `--page page_1 --page page_2`, otherwise the last argument will be assigned to the dump file `path` variable.

This format `--page page_1 page_2` should set the argument action to `nargs="*"`, but that requires the dump path passed before any `--page` argument: `enwiktionary-*** --page page_1`

kristian-clausal:
Although I don't see any reason why anyone would reasonably want to use --page without a ready database file... Actually, I can't really come up with any arguments why --page shouldn't be only available with --db-path and without the wiktionary dump file. But just in case, let's let it be an option if someone wants to use it.

I'll add some prints and error messages.

PR #256, opened 2023-06-02
Add Chinese Wiktionary thesaurus pages extractor

xxyzz:
Changes in this pull request:
- Save thesaurus data to a database file, extracting thesaurus will be skipped if the database file exists and has data.
- Move English thesaurus extractor code from `wiktextract/thesaurus.py` to `wiktextract/extractor/en/thesaurus.py`, and load different extractor according to the dump file language code.
- Add Chinese Wiktionary thesaurus pages extractor.


xxyzz:
I didn't translate the tags, for example "正式" -> "formal". Because I think these values are too broad to be categorized.

PR #265, opened 2023-06-16
Add Chinese Wiktionary extractor

xxyzz:
I have tested a few Chinese Wiktionary pages([大家](https://zh.wiktionary.org/zh/大家), [manga](https://zh.wiktionary.org/wiki/manga), [癸卯](https://zh.wiktionary.org/wiki/癸卯)), and the code can extract etymology, gloss, example and pronunciation data.

Changes:
- I changed the "english" key to "translation" in the "examples" dictionary for `zh/page.py` extractor, but didn't change the code in `en/page.py` file.
- and use a spacial page link to create the ogg file link

For future works:
- categories data aren't created in this pull request
- sense tags aren't translated, that could be done by extracting translation data from https://zh.wiktionary.org/wiki/Module:Labels/data.
- `extract_examples()`, `extract_pronunciation()` could be moved to separate files if they grow longer

### DIFF
--- a/wiktextract/clean.py
+++ b/wiktextract/clean.py
@@ -1381,8 +1381,8 @@ def clean_value(wxr, title, no_strip=False, no_html_strip=False):
     # Change <div> and </div> to newlines.  Ditto for tr, li, table, dl, ul, ol
     title = re.sub(r"(?si)</?(div|tr|li|table|dl|ul|ol)\b[^>]*>",
                    "\n", title)
-    # Change </dt> and </dd> into newlines; these generate new rows/lines.
-    title = re.sub(r"(?si)</(dt|dd)\s*>", "\n", title)
+    # Change <dt> and <dd> into newlines; these generate new rows/lines.
+    title = re.sub(r"<d[dt]>", "\n", title)
     # Change <td> </td> to spaces.  Ditto for th.
     title = re.sub(r"(?si)</?(td|th)\b[^>]*>", " ", title)
     # Change <sup> ... </sup> to ^
--- a/wiktextract/extractor/zh/page.py
+++ b/wiktextract/extractor/zh/page.py
@@ -275,8 +275,22 @@ def extract_examples(
                             else:
                                 key = "translation"
                             example_data[key] = expanded_line
-                    elif template_name == "zh-usex":
-                        pass
+                    elif template_name in {"zh-x", "zh-usex"}:
+                        for expanded_line in expanded_text.splitlines():
+                            if expanded_line.endswith("體]"):
+                                # expanded simplified or traditional Chinese
+                                # example sentence usually ends with
+                                # "繁體]" or "簡體]"
+                                if example_data.get("text") is not None:
+                                    example_data["text"].append(expanded_line)
+                                else:
+                                    example_data["text"] = [expanded_line]
+                            elif expanded_line.endswith("]"):
+                                example_data["roman"] = expanded_line
+                            elif expanded_line.startswith("來自："):
+                                example_data["ref"] = expanded_line[3:]
+                            else:
+                                example_data["translation"] = expanded_line
                     else:
                         example_data["text"] = expanded_text
 
--- a/wiktextract/extractor/zh/page.py
+++ b/wiktextract/extractor/zh/page.py
@@ -1,0 +1,293 @@
+import copy
+import logging
+import re
+import string
+
+from typing import Dict, List, Union, Any
+
+from wikitextprocessor import WikiNode, NodeKind
+from wiktextract.wxr_context import WiktextractContext
+from wiktextract.page import clean_node, LEVEL_KINDS
+
+
+# Templates that are used to form panels on pages and that
+# should be ignored in various positions
+PANEL_TEMPLATES = {
+    "CJKV",
+    "French personal pronouns",
+    "French possessive adjectives",
+    "French possessive pronouns",
+    "Han etym",
+    "Japanese demonstratives",
+    "Latn-script",
+    "Webster 1913",
+    "attention",
+    "attn",
+    "character info",
+    "character info/new",
+    "character info/var",
+    "delete",
+    "dial syn",
+    "dialect synonyms",
+    "examples",
+    "hu-corr",
+    "hu-suff-pron",
+    "interwiktionary",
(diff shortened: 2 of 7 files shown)


wiktextract/extractor/zh/page.py
```diff
@@ -0,0 +1,293 @@
...
+    )
+
+    page_data = []
+    for node in filter(
+        # ignore link created by `also` template at the page top
+        lambda n: isinstance(n, WikiNode) and n.kind != NodeKind.LINK,
+        tree.children,
+    ):
+        if node.kind != NodeKind.LEVEL2:
+            logging.warning(f"Unexpected top-level node: {node}")
```
kristian-clausal, line 479: Should migrate to Wtp.debug(), .warning() and .error() for compatibility with our other logging functions.

kristian-clausal (approved):
Looks good for bare-bones implementation :+1:

kristian-clausal:
Ah, I closed my review too early by mistake. I'll look through the other commits quickly, but I don't think there should be anything problematic here (that I would be able to spot), and it's for the zh-edition parsing so shouldn't affect en-edition parsing.

wiktextract/extractor/zh/page.py
```diff
@@ -323,17 +323,30 @@ def split_pronunciation_tags(text: str) -> List[str]:
...
-    # breakpoint()
     if expanded_text.startswith("File:"):
         return expanded_text
     else:
-        sound_tags, ipa = re.split("：|:", expanded_text, 1)
-        data = {"tags": list(set(tags + split_pronunciation_tags(sound_tags)))}
-        ipa_key = "zh-pron" if lang_code == "zh" else "ipa"
-        data[ipa_key] = ipa.strip()
-        return data
+        sound_tags, *ipa = re.split("：|:", expanded_text, 1)
```
kristian-clausal, line 431: Using an r""-string is probably best for when someone needs to edit this to add something that has been escaped and they don't realize the string isn't an r-string.

wiktextract/clean.py
```diff
@@ -1381,8 +1381,8 @@ def repl_1_syntaxhighlight(m):
     # Change <div> and </div> to newlines.  Ditto for tr, li, table, dl, ul, ol
     title = re.sub(r"(?si)</?(div|tr|li|table|dl|ul|ol)\b[^>]*>",
                    "\n", title)
-    # Change </dt> and </dd> into newlines; these generate new rows/lines.
-    title = re.sub(r"(?si)</(dt|dd)\s*>", "\n", title)
+    # Change <dt> and <dd> into newlines; these generate new rows/lines.
+    title = re.sub(r"<d[dt]>", "\n", title)
```
kristian-clausal, line 1385: The whitespace at the end of a HTML tag is valid HTML (not at the start), so we have to keep the `\s*` in case someone does type `<dt >`.

kristian-clausal: I think both the start and end tags could be turned into newlines, just in case. `r"</?d[dt]\s*>`

kristian-clausal (approved):

wiktextract/clean.py
kristian-clausal: I'll just do it quick-like.

wiktextract/extractor/zh/page.py
```diff
@@ -0,0 +1,293 @@
...
+    "langhd",
+}
+
+
+def append_dict(
+    page_data: List[Dict], field: str, value: Any, base_data: Dict
+) -> bool:
+    if (
+        page_data[-1].get(field) is not None
+        and len(page_data[-1]["senses"]) > 0
```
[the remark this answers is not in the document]
xxyzz: I use this function to add the top level key value like "pos", "etymology_text" and "sounds", if the last dictionary in the main list has value for key "senses" I'd assume the data for that block are already added then append a new base dictionary to the main list.

I have read the English page.py and it uses different dictionaries(`pos_data`, `pos_datas`, `sense_data`, etc.) to save data, I feel like that's more complicated so I save the data to the final list directly.

wiktextract/extractor/zh/page.py
```diff
@@ -0,0 +1,293 @@
...
+            ],
+        )
+        extract_sense_tags(wxr, page_data, gloss)
+
+
+def extract_sense_tags(
+    wxr: WiktextractContext, page_data: List[Dict], raw_gloss: str
+) -> None:
+    if raw_gloss.startswith("("):
+        label_end = raw_gloss.find(")")
```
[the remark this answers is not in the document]
xxyzz: I didn't think of this case... and I have certainly missed out many cases, I guess we could modify the relevant code from the English extractor and generalize it.

wiktextract/extractor/zh/page.py
```diff
@@ -0,0 +1,293 @@
...
+            etymology = clean_node(
+                wxr,
+                None,
+                [
+                    child
+                    for child in node.children
+                    if not isinstance(child, WikiNode)
+                    or (
+                        child.kind != NodeKind.LIST
+                        and child.kind not in LEVEL_KINDS
```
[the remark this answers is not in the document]
xxyzz: Unlikely but it's children could contain level node like pos subtitle, my later commits find the first level node in the etymology title node's children and only extract etymology text before that level node.

wiktextract/extractor/zh/page.py
xxyzz: Sure, I'll change them. I guess the `sortid` argument of `Wtp.error()` is used to group and sort errors on kaikki.org? So it's OK to use outdated line number as long as it's unique?

wiktextract/extractor/zh/page.py
xxyzz: Thanks, I somehow forget to use the raw string for regex pattern at here.

wiktextract/clean.py
xxyzz: Thanks!

wiktextract/extractor/zh/page.py
kristian-clausal: That is correct, it's just a handy way to sort them. Originally, I used a costly function to check the traceback and find where the logging call was *from*, but that was too expensive so I just added manual string parameters instead.

PR #279, opened 2023-07-10
Extract form and gender data from Chinese Wiktionary headword line

xxyzz:
`assert` are removed from `data_append()` and `data_extend()` because `Mock` objects won't pass `assert isinstance`.

### DIFF
--- a/tests/test_zh_headword.py
+++ b/tests/test_zh_headword.py
@@ -1,0 +1,63 @@
+from unittest import TestCase
+from unittest.mock import patch, Mock
+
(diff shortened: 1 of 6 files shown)


kristian-clausal:
https://docs.python.org/3/library/unittest.mock.html#the-mock-class

Have you tried out `spec=Wtp()` yet? If `spec` is an object, `__class__` (used by isinstance) should apparently return the class of the object.

If this fails, please attempt to rewrite the test with an instantiated Wtp if needed. Tatu prefers to have the asserts in place.

PR #280, opened 2023-07-11
Prase expanded headword line HTML tags

xxyzz:
Check HTML tag and class name to find form text, gender text and romanization text, this is more reliable then only process the expanded plain text.

The mock HTML return value in tests are written in a single line because I tried to format them but `ikitextprocessor.Wtp.parse()` adds some extra `preformated` nodes.


kristian-clausal:
I thought I had already merged this! Probably forgot waiting for tests to pass.

PR #281, opened 2023-07-12
Fix insert thesaurus data concurrency bug

xxyzz:
`insert_thesaurus_term()` runs some queries before committing, which causes race contion when called from multiple processes.


kristian-clausal:
Looks good to me, wondered about removing `subtitle_tags = ()` but of course if you return immediately after that it doesn't matter.

xxyzz:
The `subtitle_tags` variable could be removed, it's not used in the code.

PR #283, opened 2023-07-14
Some changes to the Chinese Wiktionary extractor

xxyzz:
I want to add a test for the ja-i template but haven't figure out a simple way to create a expanded table node tree.


kristian-clausal:
I'm at the office for a hot minute to try to change how the kaikki regeneration log output looks in the e-mails it sends, so I'll just quickly merge this.

PR #286, opened 2023-07-17
Ignore templates that are inside variable `stop_head_at_these_templates`

### LINKED ISSUE #285: Missing definition for some English words
There appears to be 331 words without a definition.  A good example to look at is *ox*, which has a definition on [wiktionary](https://en.wiktionary.org/wiki/ox), but not in the extracted json files.

xxyzz:
If we break at the category template, the gloss list nodes after it won't be processed. Fixes #285.

### DIFF
--- a/tests/test_alt_form_of.py
+++ b/tests/test_alt_form_of.py
@@ -3,18 +3,25 @@
 # Copyright (c) 2021-2022 Tatu Ylonen.  See file LICENSE and https://ylonen.org
 
 import unittest
-from wiktextract.wxr_context import WiktextractContext
+
 from wikitextprocessor import Wtp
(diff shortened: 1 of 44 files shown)


xxyzz:
I believe the `node` variable at here is incorrect(line 1074): https://github.com/tatuylonen/wiktextract/blob/e231f832fdb63e811595416e21799feebe419bcb/wiktextract/extractor/en/page.py#L1070-L1113

because it was defined in the previous `for` loop: https://github.com/tatuylonen/wiktextract/blob/e231f832fdb63e811595416e21799feebe419bcb/wiktextract/extractor/en/page.py#L969

But I don't know what this code is trying to do...

PR #315, opened 2023-08-14
Fix sqlite error

xxyzz:
I'm confused about this sqlite error, if it's about the connection can't be used in forked worker process then I should see the same error of the `Wtp.db_conn`. Maybe it's because how these variables are passed.

`Wtp` is copied to a global variable `_global_ctx` in `wikitextprocessor.core.py`: https://github.com/tatuylonen/wikitextprocessor/blob/ba5e92f3972dff924d142f470088590ae0a2b9c8/wikitextprocessor/core.py#L1835-L1837

but `WiktextractContext` is passed in a callback function : https://github.com/tatuylonen/wiktextract/blob/c9567026e1d12f694c15b1676c22bcaf7949ee1b/wiktextract/wiktionary.py#L122-L123

### DIFF
--- a/wiktextract/extractor/zh/pronunciation.py
+++ b/wiktextract/extractor/zh/pronunciation.py
@@ -4,6 +4,7 @@ from typing import Dict, List, Optional, Union
 
 from wikitextprocessor import NodeKind, WikiNode
 
+from wiktextract.datautils import append_base_data
 from wiktextract.extractor.share import WIKIMEDIA_COMMONS_URL, contains_list
 from wiktextract.page import clean_node
 from wiktextract.wxr_context import WiktextractContext
@@ -17,8 +18,6 @@ def extract_pronunciation_recursively(
     node: Union[WikiNode, List[Union[WikiNode, str]]],
     tags: List[str],
 ) -> None:
-    from .page import append_page_data
-
     if isinstance(node, list):
         for x in node:
             extract_pronunciation_recursively(
@@ -58,7 +57,7 @@ def extract_pronunciation_recursively(
                         create_audio_url_dict(data)
                     )
         elif isinstance(data, dict):
-            append_page_data(
+            append_base_data(
                 page_data,
                 "sounds",
                 [data],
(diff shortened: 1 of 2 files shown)


kristian-clausal:
We had a power outage which interrupted a kaikki regen run I started this morning to see if I could recreate the deadlocks, but I'm going to wait for this and have it run after work.

xxyzz:
I can confirm the variables in callback function is not properly copied:

```python
from multiprocessing import Pool

A = []

def myfunc(x):
    print(f'mycallback is called with {x}')
    A.append(x)
    print(f"{A=} {x=}")


with Pool() as pool:
    for r in pool.map(myfunc, range(10)):
        pass

print(f"main process: {A=}")
``` 

has this output:
```
mycallback is called with 0
A=[0] x=0
mycallback is called with 1
A=[1] x=1
mycallback is called with 3
A=[0, 3] x=3
mycallback is called with 4
A=[0, 3, 4] x=4
mycallback is called with 2
A=[2] x=2
mycallback is called with 5
A=[5] x=5
mycallback is called with 7
A=[7] x=7
mycallback is called with 8
A=[8] x=8
mycallback is called with 6
A=[6] x=6
mycallback is called with 9
A=[9] x=9
main process: A=[]
```

we really shouldn't rely on the `fork` behavior on Linux.

kristian-clausal:
Yeah, if we are doing something like that, then it's wrong.

xxyzz:
Unfortunately, that's exactly how variables in most callback functions to worker processes in both projects. But since we have most features and stats variables inside a single class, that would be tricky to separate.

And I'm thinking if MediaWiki has this [HTML spec](https://www.mediawiki.org/wiki/Specs/HTML) then why bother to expand and parse wikitext?

xxyzz:
The conclusion I posted above is wrong, list `A` is not shared among worker processes, it looks that way because some worker tasks are assigned to the same pool process. And the sqlite error might be related to process hung at `decode_tags()`.

PR #335, opened 2023-09-13
Fix errors in extract French Wiktionary inflection table code

xxyzz:


xxyzz:
I have fixed all the Python exceptions in the French Wiktionary extractor except errors in thesaurus pages(because the English Wiktionary thesaurus extractor is used), and the code now takes half hour to create a 3GB JSON file.

PR #336, opened 2023-09-15
Fix deploy page action timeout error

xxyzz:
The 20M "taxondata.py" file creates a ridiculous 143M coverage HTML file, which makes the deploy action failed after the default 10m timeout.


kristian-clausal:
Recently I added newlines to taxondata, because trying to read a single-line 20mb text file in any editor I tried was problematic. Hopefully that's not what caused this.

xxyzz:
That HTML file has always been 143M, it's GitHub deploy action becomes too slow. IMO we shouldn't use huge text file in the first place...

PR #338, opened 2023-09-17
Fix 'parse_page' for French extractor

empiriker:
While exploring the progress to the French extractor you've made recently (well, done already!), I wanted to run the script from the default entry point via `wiktwords.py` but quickly discovered that there are still problems in `fr/page.py`.

One I fixed already: 
- https://github.com/tatuylonen/wiktextract/commit/fd53cdd1dfc5b3fc2a84595ab23cb113525ca1d5: Originally, empty dicts were created  instead of empty sets.

The main problem: 
- In several places, the code makes use of the inexistent instance method `find_content` on objects of the `WikiNode` class, for example in https://github.com/tatuylonen/wiktextract/blob/05321e53017a175dd2c7a293e1b4f784764a81f7/wiktextract/extractor/fr/page.py#L217 This seems intentional but I'am unsure what to do about it. @xxyzz Does this rely on some not yet merged feature?

For easier testing, I've added a simple test suite for the `parse_page` function of the French extractor.

ToDo:
- [x] replace or implement non existent `Wikinode.find_content()` method
- [x] run test suite successfully

### DIFF
--- a/tests/test_fr_page.py
+++ b/tests/test_fr_page.py
@@ -1,0 +1,58 @@
+# Tests for parsing a page
+#
+# Copyright (c) 2021 Tatu Ylonen.  See file LICENSE and https://ylonen.org
+
+import unittest
+
+from wikitextprocessor import Wtp
+
+from wiktextract.config import WiktionaryConfig
+from wiktextract.extractor.fr.page import parse_page
+from wiktextract.thesaurus import close_thesaurus_db
+from wiktextract.wxr_context import WiktextractContext
+
+
+class FrPageTests(unittest.TestCase):
+    def setUp(self):
+        self.maxDiff = None
+        conf1 = WiktionaryConfig(
+            dump_file_lang_code="fr",
+            capture_language_codes=None,
+            capture_translations=True,
+            capture_pronunciation=True,
+            capture_linkages=True,
+            capture_compounds=True,
+            capture_redirects=True,
+            capture_examples=True,
+        )
+        self.wxr = WiktextractContext(Wtp(lang_code="fr"), conf1)
+
+    def tearDown(self) -> None:
+        self.wxr.wtp.close_db_conn()
+        close_thesaurus_db(
+            self.wxr.thesaurus_db_path, self.wxr.thesaurus_db_conn
+        )
+
+    def test_fr_parse_page(self):
+        self.wxr.wtp.add_page("Modèle:langue", 10, "Français")
+        self.wxr.wtp.add_page("Modèle:S", 10, "Nom commun")
+        lst = parse_page(
+            self.wxr,
+            "exemple",
+            """
+== {{langue|fr}} ==
+=== {{S|nom|fr}} ===
+'''exemple'''
+""",
+        )
+        self.assertEqual(
+            lst,
+            [
+                {
+                    "lang": "Français",
+                    "lang_code": "fr",
+                    "pos": "noun",
+                    "word": "exemple",
+                }
+            ],
+        )
--- a/wiktextract/extractor/fr/page.py
+++ b/wiktextract/extractor/fr/page.py
@@ -19,15 +19,15 @@ from .translation import extract_translation
 
 # Templates that are used to form panels on pages and that
 # should be ignored in various positions
-PANEL_TEMPLATES = {}
+PANEL_TEMPLATES = set()
 
 # Template name prefixes used for language-specific panel templates (i.e.,
 # templates that create side boxes or notice boxes or that should generally
 # be ignored).
-PANEL_PREFIXES = {}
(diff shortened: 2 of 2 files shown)


xxyzz:
`find_content()` is the method of `LevelNode` class: https://github.com/tatuylonen/wikitextprocessor/blob/main/wikitextprocessor/parser.py#L541

Those empty variables are just for compatible with old code in `page.py` and other files, they are not used for now.

Each subtitle sections are tested separately in their own test file, I think it's no harm to add a simple test for `parse_page()` but it shouldn't duplicate the existing tests.

tests/test_fr_page.py
```diff
@@ -0,0 +1,73 @@
...
+        self.wxr.wtp.close_db_conn()
+        close_thesaurus_db(
+            self.wxr.thesaurus_db_path, self.wxr.thesaurus_db_conn
+        )
+
+    def runpage(self, text):
+        assert isinstance(text, str)
+        return parse_page(self.wxr, self.wxr.wtp.title, text)
+
+    def test_page1(self):
```
xxyzz, line 43: I guess you copied the code from `tests/test_page.py` file, that file is not a good example... I have updated it: https://github.com/tatuylonen/wiktextract/commit/098461c5b1b60163fc19b603bbf67b2fc0de9eb6

And please use meaning name for test functions.

empiriker: Thanks for the review.

Yes, I didn't yet consider the final test form. I just wanted something quickly to run the parse_page function. Will improve!

empiriker:
> `find_content()` is the method of `LevelNode` class: https://github.com/tatuylonen/wikitextprocessor/blob/main/wikitextprocessor/parser.py#L541
> 

@xxyzz Thanks for the pointer. That led me on the right path. My wikitextprocessor dependency was not up-to-date even though I ran the wiktextract installation again (`python -m pip install --use-pep517 -e ".[dev]"`). Since the version number of wikitextprocessor doesn't change it was satisfied with the old install. 

What's the best practice then to keep both libraries in sync (apart from deleting the python environment and reinstalling everything from scratch)?

xxyzz:
I'd recommend install wiktextract first then force reinstall wikitextprocessor, both with the `-e` option and in the same venv.

empiriker:
@xxyzz Just to clarify: You mean reinstall wikitextprocessor with the -e option from a local clone? Then I'd should just have to pull the most recent changes periodically.

xxyzz:
Yes, clone both projects to your machine and install them with the `-e`(editable) option.

tests/test_fr_page.py
```diff
@@ -0,0 +1,74 @@
...
+        close_thesaurus_db(
+            self.wxr.thesaurus_db_path, self.wxr.thesaurus_db_conn
+        )
+
+    def mock_clean_node(self, *args, **kwargs):
+        try:
+            if args[2].template_name == "langue":
+                return "Français"
+        except:
+            return "Nom commun"
```
xxyzz, line 42: You could add these two templates using `Wtp.add_page()` and replace the page body with plain text(don't call lua code) before calling `parse_page()`.

tests/test_fr_page.py
```diff
@@ -0,0 +1,74 @@
...
+            self.wxr,
+            "exemple",
+            """
+== {{langue|fr}} ==
+=== {{S|nom|fr}} ===
+'''exemple'''
+""",
+        )
+        self.assertEqual(
+            [dict(d) for d in lst],  # Convert defaultdicts to dicts
```
xxyzz, line 65: No need to convert `defaultdicts`.

empiriker:
I've updated the test file and now with some mocking and an up-to-date version of wikitextprocessor everything works.

The test doesn't cover much apart from that the parse_page function can be called correctly. 

That at least can catch error such as fixed by https://github.com/tatuylonen/wiktextract/pull/338/commits/d4540334051ba5395210d9aac57ac31f9a1624cd. Without properly declaring empty sets the assertion
` assert additional_expand is None or isinstance(additional_expand, set)`
[(Line 1852)](https://github.com/tatuylonen/wikitextprocessor/blob/9ac05aad22a5a828fd40f5ab5c29b172a328587b/wikitextprocessor/core.py#L1852)
would fail.

tests/test_fr_page.py
empiriker: Thanks!

tests/test_fr_page.py
empiriker: Yes, that's more elegant.

empiriker:
By the way, do you prefer it that I use force-push (as is did so far) or that I create new commits while responding to your code reviews?

xxyzz:
Force push is fine. And thanks for your contribution!

PR #342, opened 2023-09-20
Add German Wiktionary extractor code to parse page and extract glosses

empiriker:
Hi there!

After the great restructuring you did over the summer to support and maintain different extractors for different Wiktionary editions, I thought to jump right in, try out this new framework and start building an extractor for the German Wiktionary.

I will try to keep each pull request small but complete. Is that in your interest?

### General questions

I have modelled my code on the French extractor and tried to follow the general style of the code base. In this regard maybe two preliminary questions:
1. Do you still use the # XXX flag to mark todos/stuff that isn't properly extracted yet?
2. Is there any guidance on when to log a _warning_ vs. a _debug_ statement? What is the `sortid` supposed to look like (especially the number at the end)?


kristian-clausal:
Use #XXX for anything that isn't yet finished, it's just the convention in the codebase.

debugs and warnings are... I don't actually use warnings myself. Debugs is for things that could be correct, but possibly isn't, for printing out stuff you need to know for debugging, the biggest category by far. I guess warnings are for more explicit "this shouldn't be like this" stuff, possibly caused by stuff on wiktionary-side that isn't our own bug. But it's vague, yeah.

The sortid is just something I threw in there so that the log files have the same error kind next to each other when sorting, so it's just an arbitrary string. I usually put the file name, line number (approximate, doesn't really matter) and recently also the ISO date to really prevent collisions.

wiktextract/extractor/de/page.py
```diff
@@ -0,0 +1,508 @@
...
+    "Latein Adjektiv Übersicht",
+    "Latein Adverb Übersicht",
+    "Latein Substantiv Übersicht",
+    "Lettisch Substantiv Übersicht",
+    "Lettisch Verb Übersicht",
+    "Mazedonisch Substantiv Übersicht",
+    "Nahuatl Substantiv Übersicht",
+    "Neugriechisch Substantiv Übersicht",
+    "Niederdeutsch Adjektiv Übersicht",
+    "Niederländisch Adjektiv Übersicht" "Niederländisch Substantiv Übersicht",
```
xxyzz, line 100: Missing a comma here.

wiktextract/extractor/de/page.py
```diff
@@ -0,0 +1,508 @@
...
+# Templates that should not be pre-expanded
+DO_NOT_PRE_EXPAND_TEMPLATES = {
+    "Ü-Tabelle",  # Translation table
+    "Quellen",  # Can be ignored since we have the <ref> tags in the tree
+}
+
+DO_NOT_PRE_EXPAND_TEMPLATES.update(FORM_TABLE_TEMPLATES)
+
+
+def fix_level_hierarchy_of_subsections(
```
xxyzz, line 156: I think overwrite the subtitle template pages would be easier, like the [overrides/zh.json](https://github.com/tatuylonen/wiktextract/blob/master/overrides/zh.json) file

empiriker:
I have moved the list of German form table templates to a different file. That makes it a little clearer, I think.

I think it will be useful to have this comprehensive list of templates containing form data if we choose to extract it later. Though there is some argument to be made that the English Wiktionary edition most likely already includes all this form data. This means it deserves some investigation whether or not extracting form data from each Wiktionary edition is warranted.

empiriker:
Since there was no further comment, I decided to keep the test for `parse_section()` that makes the most sense to me.

@xxyzz Any objection to merging this pull request?

tests/test_de_page.py
```diff
@@ -0,0 +1,213 @@
...
+    def test_de_parse_section(self):
+        self.wxr.wtp.add_page("Vorlage:Wortart", 10, "")
+        self.wxr.wtp.add_page("Vorlage:Bedeutungen", 10, "")
+        page_text = """
+=== {{Wortart|Adjektiv|Englisch}}, {{Wortart|Adverb|Englisch}} ===
+{{Bedeutungen}}
+:[1] gloss1
+=== {{Wortart|Verb|Englisch}} ===
+{{Bedeutungen}}
+:[1] gloss1
```
kristian-clausal, line 95: The data here should be different from data elsewhere, otherwise your tests might pass if there's cross-contamination; if glossAdverb is accidentally used for Verb, when it should be glossVerb, this doesn't catch that. So gloss1, gloss1, gloss1 > gloss1, gloss2, gloss3, etc.

kristian-clausal:
xxyzz is on vacation right now, but he'll be back at some point. I haven't kept up with the stuff happening in these extractors because I've been hitting my head on the wall trying to fix elusive, non-deterministic Lua bugs, but I had a small comment for one of the tests: when you use "gloss1" for each gloss in separate sections, you can't actually tell if the test fails in cases where the data from one section is accidentally contaminating another. This *is* unlikely, but possible if something goes spectacularly wrong. Otherwise looks fine, what I saw; de.wiktionary has different testing needs, as do all other editions, too, so all we can do is trust that those who implement the extractors know those idiosyncrasies and differences best.

empiriker:
@kristian-clausal Fair point. And that actually solves my philosophical problem with the test. :smile:  While still including the `extract_glosses` function in its test scope, it at least tests something that unittesting `extract_glosses` will never test: How its results get distributed to the different page_data items. 

Alright. Then let's wait for xxyzz to get back.

kristian-clausal:
The commits do not seem to touch anything outside of the de.wiktionary extractor stuff, so as long as the tests pass I see no problem merging if you feel like you've reached a milestone.

wiktextract/extractor/de/page.py
xxyzz: wiktextract extractor code usually should not change the node tree structure, and I think for the German Wiktionary my above comment is simpler and more reliable. For example, we could override subtitle templates: `{"Vorlage:Beispiele": {"pre_expand": true, "body": "==== Beispiele ====", "namespace_id": 10}}`.

wiktextract/data/de/form_tables.json
```diff
@@ -0,0 +1,119 @@
+[
+  "Deutsch Substantiv Übersicht",
```
xxyzz, line 2: I'd prefer not including lots of hard coded data at the beginning, I think this long template list is unnecessary for these reasons:
- most can be found by a simple `template_name.endswith("Übersicht")` 
- and all these table templates won't be pre-expanded thus won't needed to be included in the `do_not_pre_expand` parameter. If they are pre-expaned then the problem is `Wtp.analyze_templates()` function marks too many pre-expand templates and should only be called by the English Wiktionary code and in the future we should make the pre-expand rules configurable like only check if the template contains list and ignore other cases.

I'd recommend you also take a look of the newer French extractor code which is simpler than the English extractor and doesn't use much hard code data and regex.

usertools/de_language_data.py
```diff
@@ -0,0 +1,72 @@
+# Export German Wiktionary language data to JSON.
```
xxyzz, line 1: Languages JSON files are created from the code at here: https://github.com/tatuylonen/wiktextract/tree/master/languages

This code should be moved there called from the `get_data.py` file.

PR #346, opened 2023-09-25
Extract French Wiktionary etymology list

xxyzz:
This pull request extracts French Wiktionary etymology list and breaks the list text to JSON list "etymology_texts"(this is a str type for English JSON file), examples can be found in test cases.

### DIFF
--- a/tests/test_fr_etymology.py
+++ b/tests/test_fr_etymology.py
@@ -4,7 +4,10 @@ from collections import defaultdict
 from wikitextprocessor import Wtp
 
 from wiktextract.config import WiktionaryConfig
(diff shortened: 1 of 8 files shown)


kristian-clausal:
At some points it would be good to go through all the differences that have accumulated between the en and fr/zh versions of the data 'scheme', so that we can consolidate them (probably from en -> fr/zh).

xxyzz:
This pr also adds a new "pos_title" in French JSON for matching etymology text. fr/zh uses "translation" in the "examples" list(it is "english" in English JSON). And Chinese JSON breaks Simplified/Transitional character example sentences to a list(also str in English JSON). That's some differences I can remember for now.

PR #353, opened 2023-10-07
Simplify `parse_page()` in extractor code

xxyzz:
Only the level two language title node is used in these functions, other nodes are ignored and the code could be simplified with the `WikiNode.find_child()` method.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.10', '3.11']
+        python-version: ['3.9', '3.10', '3.11', '3.12']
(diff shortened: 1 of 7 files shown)


empiriker:
I see some changes to the import order in the /de files.

Do you use some kind of auto-sort on the imports? I'd be happy to follow the prefered order.

xxyzz:
I used the `isort` tool, it's inside the `dev` dependency group.

xxyzz:
Hi @empiriker, I left some comments to your German extractor [pull request](https://github.com/tatuylonen/wiktextract/pull/342) earlier today, please check them out.

xxyzz:
I think all the `invalid escape sequence` syntax warnings in the 3.12 test outputs are regex bugs...

PR #356, opened 2023-10-09
Add French and Chinese Wiktionary JSON schema

xxyzz:
I'm run the validation code but it is not finish yet, the schema might not match the JSON file.

### DIFF
--- a/json_schema/fr.json
+++ b/json_schema/fr.json
@@ -131,6 +131,12 @@
     "redirect": {
       "description": "Redirect page target title",
       "type": "string"
+    },
+    "categories": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
     }
   },
   "$defs": {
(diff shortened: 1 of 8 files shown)


xxyzz:
The script is finished successfully. But it takes 30 minutes, almost the same as creating the JSON file.

kristian-clausal:
Tatu wants that all of the json data produced by wiktextract should be as close to each other as possible, so any schemas should probably be also applied to all other extraction output. We will have to consolidate at some point (when we create kaikki.org dictionaries for the French and Chinese wiktionaries etc. at some point), which probably means learning from these newer schemas and applying changes to the en.wiktionary output, and then checking for divergences that might have slipped in by accident.

xxyzz:
The main structures are the same as the English JSON data, but both French and Chinese Wiktionary have their own unique extra data, like French Wiktionary has traditional writing form in translation list and Chinese Wiktionary has two forms of example sentences. Some parts of the schemas might be incompatible, but the validation script would still pass because I didn't add any required property.

xxyzz:
Chinese Wiktionary JSON schema validation also passed.

PR #361, opened 2023-10-13
Fix regex alerts

xxyzz:
This pr fixes all alerts in GitHub Code Scanning exception one alert of this regex pattern: https://github.com/tatuylonen/wiktextract/blob/08407b3a8031ec31482f2b457789e0896768ad07/src/wiktextract/clean.py#L1446

I don't know how to fix that pattern...


kristian-clausal:
The corrections seem correct. :+1:

PR #363, opened 2023-10-13
Extract categories and subglosses, examples with references and pronunciation data from German Wiktionary

empiriker:
The PR got quite large despite my best intentions. But I accumulated some progress while the last PR was open.

---
Apart from adding the extractor functions, this pull request deals with some left over housekeeping from last time. xxyzz left some remarks that I attempted to address:

1. **Not to change the tree structure but to use the override templates functionalities.**

This is done.

2. **Move the language data script to /languages**

I moved it over there but couldn't quite figure out how to integrate it in the existing script. The problem is that I couldn't find any similar central module that contains all the language data in the German Wiktionary as exists in the French and English ones. As far as I can tell there isn't even another page that centrally defines all languages and their codes. Hence, I added the missing ones (i.e. languages that were still referenced in the different pages) manually by looking up the language tags from other sources. 

It's not ideal but the language list of this commit is pretty comprehensive and should do the job. 

3. **Avoid hard-coded list of  DE_FORM_TABLES**

I saw your recent PR addressing this and decided to leave this be until your request is merged.

---

Apart from that the scripts for the new sections should speak for themselves.

PS: I in the commit messages I have included an attribution note to the project that funds my work. I hope that is acceptable. If not, please advise whether there is another way that is acceptable to you for leaving a small trail.


src/wiktextract/extractor/de/example.py
```diff
@@ -0,0 +1,93 @@
...
+            sortid="extractor/de/examples/extract_examples/33",
+        )
+
+
+def extract_reference(
+    wxr: WiktextractContext, example_data: Dict[str, str], ref_node: WikiNode
+):
+    reference_data = defaultdict()
+
+    reference_data["raw_ref"] = clean_node(wxr, {}, ref_node.children)
```
empiriker, line 65: This line here is responsible for 90% of my debug statements. Yet, the debugging is done in `core/1021`. The all have message similar to this:

```Kaugummi/de_noun/Beispiele: DEBUG: too many args (3) in argument reference: ('Sammelwerk', '', '') at ['Kaugummi', 'Literatur']```

As far as I can tell this has nothing to do with the code I wrote but is somehow caused by the template definitions themselves. References are pretty standardized in the German Wiktionary and they all at one point call a template called "[Vorlage:Literatur](https://de.wiktionary.org/w/index.php?title=Vorlage:Literatur&action=edit)" that has a variable called Sammelwerk. 

I don't quite understand what's going on here. Do you perhaps have any suggestions what I could investigate?

Otherwise, I tend to ignore this issue since everything seems to work apart from the debug statement.

File: src/wiktextract/extractor/de/pronunciation.py [hunk 2]
```diff
@@ -0,0 +1,201 @@
...
+
+def process_ipa(
+    wxr: WiktextractContext,
+    sound_data: List[Dict],
+    nodes: List[Union[WikiNode, str]],
+):
+    if not nodes:
+        return
+
+    def process_lautschrift_template(
```
xxyzz, line 77: Please don't use nested function unless you have to, local variables in these functions are often got overwritten accidentally. And again `TemplateNode.template_patameters` should be used here.

src/wiktextract/extractor/de/pronunciation.py
```diff
@@ -0,0 +1,201 @@
...
+        append_tag(wxr, sound_data, head_node)
+    elif is_new_sound_data_entry_sep(head_node):
+        sound_data.append(defaultdict(list))
+    elif not is_empty_string(head_node):
+        wxr.wtp.debug(
+            f"Found unexpected non-Lautschrift node in IPA section: {head_node}",
+            sortid="extractor/de/pronunciation/process_ipa/57",
+        )
+
+    if nodes:
```
xxyzz, line 114: Why not write these pronunciation process functions in a `for` loop to iterate child nodes? And empty string child could be removed with the [`WikiNode.filter_empty_str_child()`](https://github.com/tatuylonen/wikitextprocessor/blob/1d9672339f903dcb003bbe06c23aadb323ff0b3f/src/wikitextprocessor/parser.py#L409) method.

tests/test_de_example.py
```diff
@@ -0,0 +1,104 @@
...
+        )
+
+    def tearDown(self) -> None:
+        self.wxr.wtp.close_db_conn()
+        close_thesaurus_db(
+            self.wxr.thesaurus_db_path, self.wxr.thesaurus_db_conn
+        )
+
+    def test_de_extract_examples(self):
+        self.wxr.wtp.start_page("")
```
xxyzz, line 28: Please also add the Wiktionary word page links or titles used in tests comments.

src/wiktextract/extractor/de/example.py
xxyzz: You can ignore this debug log. I guess it's just for checking the length of template arguments isn't longer than 2, and it should ignore the empty strings.

empiriker:
First of all, I want to say how much I appreciate the detailed reviews @xxyzz. I will now take a look one by one and incorporate your suggestions.

src/wiktextract/extractor/de/example.py
```diff
@@ -0,0 +1,93 @@
...
+            f"Found unexpected number of templates in example: {template_nodes}",
+            sortid="extractor/de/examples/extract_examples/64",
+        )
+    elif len(template_nodes) == 1:
+        template_node = template_nodes[0]
+
+        # Sometimes the title is dynamically generated from the template name,
+        # so we preset the title. If specified in the template, it will be
+        # overwritten.
+        reference_data["titel"] = template_node.largs[0][0].strip()
```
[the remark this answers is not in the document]
xxyzz: And template could be used as a template parameter, in that case `clean_node()` should be used to convert the parameter node to text.

src/wiktextract/extractor/de/example.py
empiriker: Alright. A bit annoying but otherwise not problematic.

src/wiktextract/extractor/de/example.py
```diff
@@ -0,0 +1,93 @@
...
+
+            if senseid:
+                sense_data = [
+                    sense
+                    for sense in page_data[-1]["senses"]
+                    if sense["senseid"] == senseid
+                ]
+
+                for sense in sense_data:
+                    sense["examples"].append(example_data)
```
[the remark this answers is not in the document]
empiriker: Done. Thanks!

src/wiktextract/extractor/de/pronunciation.py
empiriker: Alright. I nested the function to indicate that this it is used only here and basically just for readability. 

I unnested it now.

tests/test_de_example.py
empiriker: You mean links to the Wiktionary pages where I take example snippets from? Like this?: 
```
def test_de_extract_reference_from_literatur_template(self):
   # https://de.wiktionary.org/wiki/Beispiel```

You will see that some test cases I have are constructed. But I added references where appropriate and will try to use test cases from real Wiktionary pages going forward.

Comment (xxyzz) [reply tests/test_de_example.py:28]: Yes, but not required, just a recommendation for cases that are not easy to find the original page.

File: src/wiktextract/extractor/de/gloss.py [hunk 7]
```diff
@@ -3,58 +3,148 @@
...
+    wxr.wtp.debug(
+        f"Skipped a sense modifier in gloss list: {list_item_node}",
+        sortid="extractor/de/glosses/extract_glosses/19",
+    )
+    # XXX: We should extract the modifier. However, it seems to affect
+    # multiple glosses. Needs investigation.
+    pass
+
+
+def extract_categories_from_gloss_node(
```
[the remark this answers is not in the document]
empiriker: But generally the German Wiktionary doesn't use links to the category namespace to modify single senses. So the `clean_node()` function doesn't help here.

It mostly uses the [K](https://de.wiktionary.org/wiki/Vorlage:K) template (literally: context template) which sometimes creates an entry in a category page, sometimes just outputs information about the context of usage (which we might rather store in the "`tags`" field).

Anyway I think we need some custom logic here to capture the contextual data present in the German Wiktionary.


_As an aside: I find the `clean_node()` function a bit obscure and hard to work with. I am basically using it only to create clean text from a node. To me it would seem cleaner if all other side-effects would be moved to a separate function. But perhaps I am missing some deeper reason here._

src/wiktextract/extractor/de/pronunciation.py
xxyzz: Sorry for forgetting to mention this is also a recommendation but not a rule. I just kind of scared by the long nested functions in the English extractor...

empiriker:
I rebased to the latest commit to solve the merge conflict in `src/config.py`.

I see that I could now address the outstanding suggestion to "**Avoid hard-coded list of DE_FORM_TABLES**" but would prefer to do this in a future PR.


Do you see anything else I should resolve before the current PR can be merged?

src/wiktextract/extractor/de/gloss.py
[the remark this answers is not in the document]
xxyzz: Isn't the "K" template also expands to a category link? `clean_node` will find all the category links in the expanded text and save them to the passed object. It should work for the German Wiktionary. But if it doesn't work for some cases, you're free to use a new implantation.

I also only use it to convert node to text and save category data, I guess that's the main usage of this function. It needs better documentation...

src/wiktextract/extractor/de/pronunciation.py
empiriker: Done now. It seems I initially updated another similar case instead.

src/wiktextract/extractor/de/gloss.py
```diff
@@ -3,58 +3,167 @@
...
+            category = (prep if prep else "") + (" + " + case if case else "")
+            if category:
+                categories.append(category)
+
+            # Remove the template_node from the children of list_item_node
+            list_item_node.children = [
+                c for c in list_item_node.children if c != template_node
+            ]
+
+            gloss_data["tags"].extend(categories)
```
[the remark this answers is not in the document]
empiriker: Okay. I see. I could have really saved a lot of time here if I had had a better understanding of `clean_node()` and probably also the intermediate stage between wikicode and plain text that `wxr.wtp.node_to_html()` provides.

The problem now is that I am not quite sure how to properly write the tests. `clean_node()` now requires the templates to be present. 
- For my local tests (just to make sure everything works as I think), I gave it access to my local database but that obviously doesn't work for automatic testing. 
- Mocking seems way too complicated now with everything clean_node does.
- I could just copy the template content to the test file and add it as a page to wxr. But that could get quite out of hand depending how many subpages are needed.

Have you had this scenario before?

src/wiktextract/extractor/de/gloss.py
```diff
@@ -3,58 +3,167 @@
...
+
+
+def extract_tags_from_gloss_text(
+    gloss_data: defaultdict(list), gloss_text: str
+) -> None:
+    parts = gloss_text.split(":", 1)
```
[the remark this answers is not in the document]
empiriker: I've added some examples from real wikitext.

And it seems you're right. That might be a better strategy. However, I would rather move on at the moment since what I have works fairly well. I added an XXX comment to not forget about having another look at this.

src/wiktextract/extractor/de/gloss.py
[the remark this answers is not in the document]
xxyzz: For tests process expanded text, `Wtp.add_page` could be used to add the template page with expanded plain text or use `mock.patch` to provide a return result of the `clean_node()` function.

If a test needs to process the node tree like parsing a table, `Wtp.add_page` could be used again to add the expanded wikitext table to the template page or patch the `Wtp.node_to_wikitext` method.

Maybe breaking the complex test to some smaller tests could be easier?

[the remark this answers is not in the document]
empiriker: Okay, if that's an acceptable solution, I have manually added all pages required to be present in Wtp.

I don't think the test `test_process_K_template` itself is complicated. It might seem complicated since I added all test_cases in this one test function with `subTest()`. But each single test case is quite straightforward.

 The problem is that the K template itself is complicated (you see how much I had to add to make it work). So if we want to test at least all the examples from the K template documentation (as I do at the moment), we would have to mock quite a lot. Best to just use the original K template definition then.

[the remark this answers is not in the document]
empiriker: I simplified it anyway.

Ready to merge?

xxyzz:
Thank you for your contributions! And do you mind adding a JSON schema file for the German Wiktionary extractor? There are currently only two incomplete Chinese and French schema files in the "json_schema" folder, adding a schema file will help compare the JSON files of different extractors.

jmviz:
It seems I overreached in my assumption that all language editions treated language data similarly enough that they could all be gotten at in the same way through lua modules. Watching the recent development of the extractors for the different language editions, the amount of variation between editions is impressive!

There does seem to be [Modul:Sprache](https://de.wiktionary.org/wiki/Modul:Sprache). It appears to contain the language codes that you added manually, so it looks like it might be more authoritative than the help page. Unfortunately the codes can't be iterated over directly in lua since they are written out in tons of individual `if`s instead of placed in a lua table like the other language editions' language modules do. I wonder if the German editors might accept an edit to that Module that replaces the `if`s with a table that maps codes to names. If that were the case then the same approach as is used for the other languages would work here.

empiriker:
@jmviz That is a good observation. The [Modul:Sprache](https://de.wiktionary.org/wiki/Modul:Sprache) also has more languages than I currently added to the language.json file. (677 vs. 542) Though I took care to cover every language which has word entries.

Obviously, it would be best to have a dynamic way get the current language data and your suggestion seems the best way to do so. For now though, I will focus on fleshing out the German extractor a little more.

xxyzz:
I didn't know pydantic could create JSON schema, we could try it on the newer extractors.

PR #367, opened 2023-10-18
Add support for "capture_language_codes" in de and fr extractors

empiriker:
Currently the German and French extractors don't make use of the "capture_language_codes" configuration and capture all languages by default.

This commit adds support for this feature.


xxyzz:

kristian-clausal:
Thank you for all of this, it is very much appreciated! Every different edition of Wiktionary we get to work (even if only piecemeal) helps with later-on consolidating and improving the base (en) edition and all the others.

PR #368, opened 2023-10-18
Fix missing skip_analyze_templates argument in wiktwords.py

empiriker:
See title.


xxyzz:
Thanks! I should investigate how to use linter tools to find bugs like this in GitHub Actions.

PR #370, opened 2023-10-19
Add config.json for German extractor 

empiriker:
# Preliminary comment
I found my PRs were still too large recently. Going forward, I will try to keep each one focused on a small scope, even if that means having several open PRs in parallel. 

Please let me know if you prefer it differently.

# This PR
This PR is mainly to add a config.json and remove the hard-coded list of form-table templates for the German extractor.

See commit messages for the other minor additions.

## Question
I wonder if I should also configure the German [overrides](https://github.com/tatuylonen/wiktextract/blob/master/overrides/de.json ) to be used by default. Without them the whole extractor wouldn't work since the tree structure will be very different. Currently, the `--override` flag is not clearly enough documented for the user to understand for which editions it's mandatory.


xxyzz:
The "override" option was used for fixing broken templates and Lua modules, and for testing templates. The "overrides" folder must be moved inside the "src" folder to be included in the package, and the "wiktwords" command should apply the override JSON file automatically.  I'll implement this feature in a pr.

PR #375, opened 2023-10-20
Extract semantic relations from German Wiktionary.

empiriker:
This adds support for extracting different several kinds of semantic relations from the German Wiktionary. It covers the most prevalent way in which these sections are structured.

It does not:
- extract modifiers (tags) to a semantic relation, e.g.: in `{{Synonyme}}\n:[1] [[Kokosnusspalme]], ''wissenschaftlich:'' [[Cocos nucifera]]` from [Kokospalme](https://de.wiktionary.org/wiki/Kokospalme) the tag `wissenschaftlich` will be ignored
- extract semantic relations structured using sublists, e.g. Wortbildungen in [Byte](https://de.wiktionary.org/wiki/Byte)
- extract semantic relations formated other than via a wiki link, e.g. in `{{Redewendungen}}:[1] ''[[aller Anfang ist schwer]].` from [aller](https://de.wiktionary.org/wiki/aller) the relation will not be captured since generally italics (even with links within) seem to be used to modify a relation, not to create one

I looked into all of these cases but there just doesn't seem to be a general rule that allows clearly separating the semantic relations from the rest. 

For now, I think it's a great start.

---
### FYI

1. This is about the extent that I intend to flesh out each Wiktionary edition that I plan to cover. I hope it gives a good starting base for anyone who want to go deeper.
2. @xxyzz I would like to also capture semantic relations from the French Wiktionary. Do you mind if I take a look at that? Or do plan to implement it soon?
3. When I tackle the next Wiktionary edition (probably Spanish), I will attempt to use pydantic from the get-go. If it's easy to use, I will probably add it to the German extractor as well. Otherwise, I will just add the json schema manually.

### DIFF
--- a/src/wiktextract/extractor/de/page.py
+++ b/src/wiktextract/extractor/de/page.py
@@ -7,12 +7,12 @@ from wikitextprocessor import NodeKind, WikiNode
 from wikitextprocessor.parser import LevelNode
 
 from wiktextract.datautils import append_base_data
-from wiktextract.extractor.de.pronunciation import extract_pronunciation
 from wiktextract.wxr_context import WiktextractContext
 
 from .example import extract_examples
 from .gloss import extract_glosses
 from .pronunciation import extract_pronunciation
+from .semantic_relations import SEMANTIC_RELATIONS, extract_semantic_relations
 from .translation import extract_translation
 
 # Templates that are used to form panels on pages and that should be ignored in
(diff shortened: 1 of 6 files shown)


xxyzz:
>     2. @xxyzz I would like to also capture semantic relations from the French Wiktionary. Do you mind if I take a look at that? Or do plan to implement it soon?

French Wiuktionary's synonyms lists look very similar to the translation list, maybe could share some code. Please feel free to implement the feature.

>     3. When I tackle the next Wiktionary edition (probably Spanish), I will attempt to use pydantic from the get-go. If it's easy to use, I will probably add it to the German extractor as well. Otherwise, I will just add the json schema manually.

That would be great! Does pydantic also add comments to the schema?

empiriker:
> French Wiuktionary's synonyms lists look very similar to the translation list, maybe could share some code. Please feel free to implement the feature.

Alright. I'll take a look now.

> That would be great! Does pydantic also add comments to the schema?

It has a description field which should be mapped to a json schema comment. I will keep this in mind when testing it out.

PR #376, opened 2023-10-20
Add linkage_subtitles.json for German Wiktionary

empiriker:
After taking a closer look at the French [`linkage.py`](https://github.com/tatuylonen/wiktextract/blob/master/src/wiktextract/extractor/fr/linkage.py), I realized that I should probably follow the conventions there.

The PR:
- Adds linkage_subtitles.json for German Wiktionary
- Renames "semantic_relations" to "linkages"
- Adds support for the different `--capture_X` options
- Sorts the French `linkage_subtitles.json` alphabetically.

---
@xxyzz It seems to me that the linkage.py already captures most of the semantic relations (linkages). Is there something that I missed and still required implementation?


src/wiktextract/extractor/de/linkage.py
```diff
@@ -8,23 +8,11 @@
...
-    "Wortbildungen": "derived",
-}
 
-
-def extract_semantic_relations(
+def extract_linkages(
     wxr: WiktextractContext, page_data: List[Dict], level_node: LevelNode
 ):
-    relation_key = SEMANTIC_RELATIONS.get(level_node.largs[0][0])
+    linkage_type = wxr.config.LINKAGE_SUBTITLES[level_node.largs[0][0]]
```
xxyzz, line 15: Unless `LINKAGE_SUBTITLES` contains all the possible subtitles, this line should use `.get()`.

empiriker: Since this code is only executed after checking that the linkage_type is in `LINKAGE_SUBTITLES`, it's save. 

I am still implementing your suggestion to avoid future troubles.

empiriker:
The French `linkage.py` seems to be pretty broad and is pretty much what I intended to cover. I noted some problems where there are multiple links in one list item, e.g. `* [[laçage]] ou [[lacement]]` in the page [lacet](https://fr.wiktionary.org/wiki/lacet), but generally it does a good job.

The project that allows me to contribute to this repo is mostly interested in increasing the coverage of wiktextract for other editions while doing "well enough" in some fields (mainly glosses, examples, pronunciation, translations and linkages). So I'll have to prioritize these. I might then look to improve other bits out of personal interest. :crossed_fingers:

PR #391, opened 2023-10-31
Add languages.json for Spanish Wiktionary

empiriker:
The Spanish Wiktionary doesn't use module data to define language codes. So I had to look for other sources of language data.

Fortunately, there seems to be a comprehensive, automatically-generated list of language codes with their corresponding Spanish names in the Appendix: [Apéndice:Códigos de idioma](https://es.wiktionary.org/wiki/Ap%C3%A9ndice:C%C3%B3digos_de_idioma)

There is probably a better way of organizing the get_data logic for different languages than by just adding another script file. But for more or less a one-time thing, this should suffice.

### DIFF
--- a/languages/get_data_es.py
+++ b/languages/get_data_es.py
@@ -1,0 +1,60 @@
+# Export Spanish Wiktionary language data to JSON.
+#
+# Usage:
+#
+# python language_data.py de dewiktionary_dump_file [--languages languages_output_file]
+
+import argparse
+import json
+
+from wikitextprocessor import NodeKind, WikiNode, Wtp
+from wikitextprocessor.dumpparser import process_dump
+
+from wiktextract.config import WiktionaryConfig
+from wiktextract.page import clean_node
+from wiktextract.wxr_context import WiktextractContext
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="Export Wiktionary language data to JSON"
+    )
+    parser.add_argument("lang_code", type=str, help="Dump file language code")
+    parser.add_argument("dump", type=str, help="Wiktionary xml dump file path")
+    args = parser.parse_args()
+    wxr = WiktextractContext(Wtp(lang_code=args.lang_code), WiktionaryConfig())
+
+    wxr = WiktextractContext(
+        Wtp(
+            lang_code=args.lang_code, db_path="wikt-db_es_language_data_temp.db"
+        ),
+        WiktionaryConfig(),
+    )
+    appendix_ns_id = wxr.wtp.NAMESPACE_DATA["Appendix"]["id"]
+    process_dump(wxr.wtp, args.dump, {appendix_ns_id})
+
+    # https://es.wiktionary.org/wiki/Ap%C3%A9ndice:C%C3%B3digos_de_idioma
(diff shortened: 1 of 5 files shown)


languages/get_data_es.py
```diff
@@ -0,0 +1,60 @@
...
+    parser.add_argument("dump", type=str, help="Wiktionary xml dump file path")
+    args = parser.parse_args()
+    wxr = WiktextractContext(Wtp(lang_code=args.lang_code), WiktionaryConfig())
+
+    wxr = WiktextractContext(
+        Wtp(
+            lang_code=args.lang_code, db_path="wikt-db_es_language_data_temp.db"
+        ),
+        WiktionaryConfig(),
+    )
```
xxyzz, line 31: Shouldn't `db_path` passed from command line argument?

xxyzz:
I check the Spanish [lengua](https://es.wiktionary.org/wiki/Plantilla:lengua) template and find it calls the [getlang](https://es.wiktionary.org/wiki/Plantilla:getlang) template and expands templates that use language code as template.

For example, `{{lengua|es}}` expands `{{getlang|es}}` then expands to `{{es|texto=x}}`. They save language names to templates: https://es.wiktionary.org/wiki/Categoría:Plantillas_de_idiomas

xxyzz:
And I think if a Wiktionary edition doesn't use language code and name in Lua code then we also don't need this data.

I search the code in wikitextprocessor and find the only places use these language code are `mw.language.fetchLanguageName` and `mw.language.fetchLanguageNames` and these two functions uses the same data on all mediawiki sites. And it can be downloaded from this api: https://fr.wiktionary.org/w/api.php?action=query&meta=siteinfo&siprop=languages&format=json&formatversion=2

empiriker:
Thanks for looking into this, @xxyzz.

Let's see if I'm following you: The initial reason to pre-load the languages is so that wikitextprocessor can correctly execute modules that require these languages. So you're saying that for the Spanish Wiktionary which doesn't have modules using language data, we wouldn't need to pre-load them. Do I understand you correctly?

So let's say, we do not preload language data for these languages, then we would not have access to `wxr.config.LANGUAGES_BY_NAME` and `wxr.config.LANGUAGES_BY_CODE`. So far I have found them useful to flexibly add the language code or the language name when one of them was missing. But I guess, we could do without.

---
Regarding the [lengua](https://es.wiktionary.org/wiki/Plantilla:lengua) template. I saw it and considered it to get the language names this way. But then I would still need a definite list of all the possible language codes. And the only such list I could find was [Apéndice:Códigos de idioma](https://es.wiktionary.org/wiki/Ap%C3%A9ndice:C%C3%B3digos_de_idioma).

xxyzz:
For the wikitextprocessor package, a single JSON file downloaded from the API linked above should be enough for the `mw.language.fetchLanguageName` function. And in the wiktextract package, at least the French extractor doesn't use  language data very often, because the language code is in the template argument and the language name could be obtained by expanding the subtitle template.

English and Chinese Wiktionary use plain text language names in subtitle, they would still need the huge JSON file. But I think we could run the extract language Lua code when first extracting the dump file and save the language data to database.

PR #392, opened 2023-10-31
Add pydantic models and parse_page for Spanish Wiktionary

empiriker:
Hi there,

I started to add basic support for the Spanish Wiktionary and attempted to integrate with pydantic for data validation and schema generation.

Why I think pydantic might be a good addition to the project:
- Allows creation of dynamic json schemas directly from code (also [typescript interfaces](https://github.com/phillipdupuis/pydantic-to-typescript))
- Adds validation and (optionally) type casting to ensure consistency of our output
- Can flexibly change and adapt to different Wiktionary editions (via creating or adapting pydantic classes via inheritance)
- Replaces the current type annotations -> replaces something we already do with something that does even more

The potential downsides:
- More boilerplate code
- Performance impact (?)
- Dealing with validation errors
- Adds another dependency

At the moment, I have only setup the basic pydantic models and parse_page function for the Spanish Wiktionary. I will still need to build a little more to properly assess whether pydantic really fits the use case and is easy to work with. 

---
For now, I will mark this as a draft and look forward to hear your feedback and answer potential questions you have regarding pydantic.


src/wiktextract/extractor/es/models.py
```diff
@@ -0,0 +1,113 @@
...
+            return json_schema
+
+    with open("json_schema/es.json", "w") as f:
+        json.dump(
+            WordEntry.model_json_schema(schema_generator=JsonSchemaGenerator),
+            f,
+            indent=2,
+            ensure_ascii=False,
+            sort_keys=True,
+        )
```
empiriker, line 113: Creates a json schema from a pydantic class.

src/wiktextract/extractor/es/models.py
```diff
@@ -0,0 +1,113 @@
...
+
+if __name__ == "__main__":
+
+    class JsonSchemaGenerator(GenerateJsonSchema):
+        def generate(self, schema, mode="validation"):
+            json_schema = super().generate(schema, mode=mode)
+            json_schema["title"] = "Spanish Wiktionary"
+            json_schema["$id"] = "https://kaikki.org/es.json"
+            json_schema["$schema"] = self.schema_dialect
+            return json_schema
```
empiriker, line 104: The schema generation can be customized. Here I added some json_schema field to model the `fr.json` schema.

src/wiktextract/extractor/es/models.py
```diff
@@ -0,0 +1,113 @@
...
+    ):
+        if cls.wxr:
+            cls.wxr.wtp.debug(msg, trace=trace, sortid=sortid)
+        else:
+            logging.debug(msg)
+
+
+class BaseModelWrap(BaseModel):
+    class Config:
+        extra = "ignore"
```
empiriker, line 27: If any keys are added to a pydantic class that are not defined in the class definition, they will be ignored.

src/wiktextract/extractor/es/models.py
```diff
@@ -0,0 +1,113 @@
...
+        validate_assignment = True
```
empiriker, line 28: Assigning values to properties of a pydantic class will be validated for correct type.

src/wiktextract/extractor/es/models.py
```diff
@@ -0,0 +1,113 @@
...
+    def update(self, data: dict):
+        update = self.dict(exclude_defaults=True, exclude_none=True)
+        update.update(data)
+        for k, v in (
+            self.validate(update)
+            .dict(exclude_defaults=True, exclude_none=True)
+            .items()
+        ):
+            setattr(self, k, v)
+        return self
```
empiriker, line 39: Replicates the set update functionality but with data validation.

src/wiktextract/extractor/es/models.py
```diff
@@ -0,0 +1,113 @@
...
+            for name, value in values.items()
+            if name not in all_allowed_field_names
+        }
+        if extra_fields:
+            class_full_name = cls.__name__
+            PydanticLogger.debug(
+                msg=f"Pydantic - Got extra fields in {class_full_name}: {extra_fields}",
+                sortid="wiktextract/extractor/es/pydantic/extra_fields/33",
+            )
+        return values
```
empiriker, line 57: Since extra fields are ignored by default, we might be interested in logging when we _accidentally_ add undefined fields. This might be useful, for example, when dynamically adding key value pairs from a wiki template.

src/wiktextract/extractor/es/page.py
```diff
@@ -0,0 +1,73 @@
...
+                wxr.wtp.start_section(lang_name)
+                base_data = WordEntry(
+                    lang_name=lang_name, lang_code=lang_code, word=wxr.wtp.title
+                )
+                base_data.update(categories_and_links)
+                page_data.append(copy.deepcopy(base_data))
+                for level3_node in level2_node.find_child(NodeKind.LEVEL3):
+                    parse_section(wxr, page_data, base_data, level3_node)
+
+    return [d.model_dump(exclude_defaults=True) for d in page_data]
```
empiriker, line 73: Converts pydantic objects to dicts and excludes field that have the default value (usually empty lists or None values).

empiriker:
I have added some code comments to highlight some of the features of pydantic. Don't hesitate to ask  questions if things are unclear.

xxyzz:
I think it's fine as long as pydantic and the validation code don't get in the way and don't hurt performance. I'm kind worry about overriding Python std library's `update` method. pydantic should be able to check what data are added to the `WordEntry` class, if it can't validate `dict.update`, maybe `update` shouldn't be used.

src/wiktextract/extractor/es/models.py
xxyzz: This shouldn't be ignored, many classes in the current code already has `__slots__`. And `LoggingExtraFieldsModel` will also not needed.

empiriker:
Hi there! It took longer than expected to come back to this but the next few days should see some progress.

---
Let me comment on the outstanding pydantic questions first.

> I think it's fine as long as pydantic and the validation code don't get in the way and don't hurt performance. I'm kind worried about overriding Python std library's update method. pydantic should be able to check what data are added to the WordEntry class, if it can't validate dict.update, maybe update shouldn't be used.

I am not overriding Python's `update` method, I am simply adding a new method to the pydantic data model. In the most recent version, I could simplify a b it since, indeed, by default pydantic already validates the input when using setattr. 

I also had to add the methods `get`, `__getitem__` and `__setitem__` to the base model. This doesn't override anything but makes the pydantic classes work with the programming pattern we used when dealing with defaultdicts.

src/wiktextract/extractor/es/models.py
empiriker: I am not quite sure what's the significance of `__slots__` in this context but let me explain my reasoning for configuring the base model to ignore extra fields.`

I think there are two goals in this context:
1. The pydantic models should describe the output accurately.
2. We want to capture as much data as possible.

In the German extractor, for example, I used the following pattern to extract the references to an example from a wikicode template:
```
for key, value in template_node.template_parameters.items():
    if isinstance(key, str):
        reference_data[key.lower()] = clean_node(wxr, {}, value)
```
In this example, the expected keys aren't defined anywhere in our code but rely implicitly on the arguments of that particular template. In this particular case, that makes a lot of sense since the reference template is quite detailed and exactly what we would like to capture.

However, when defining the json schema or using pydantic without the option `extra = "ignore"`, there is always the risk that our output has keys that aren't documented anywhere. That might not be a problem in and of itself but certainly makes this data a lot less valuable. We also wouldn't satisfy goal 1.

With the option `extra = "ignore"`, only keys that are defined in the pydantic model will be output and all others suppressed. This does satisfy goal 1.

However, we might still like to capture these keys (goal 2). For this reason, I am logging all extra fields that our code tries to assign but aren't defined in the model. This will enable us to make a decision on whether or not to include this key in our model. 

In short: The combination of ignoring extra fields while logging these in a debug statement will force us to keep the pydantic models (and thus our output documentation) in sync with what we do in our code. I think this is the main benefit of using pydantic over manually maintaining json schemas (as currently done for the French extractor).

xxyzz:
Nodes after ":" are saved to the [`WikiNode.definition`](https://github.com/tatuylonen/wikitextprocessor/blob/de5b4c77014c7bb9e1f710a00c8779478e3ac5c3/src/wikitextprocessor/parser.py#L335) attribute. It's created at here: https://github.com/tatuylonen/wikitextprocessor/blob/de5b4c77014c7bb9e1f710a00c8779478e3ac5c3/src/wikitextprocessor/parser.py#L694-L705

empiriker:
Thanks, @xxyzz. I haven't worked with the `WikiNode.definition` before. Always great to learn something new!

I have added gloss extraction now and so far am quite happy with using pydantics to build the Spanish extractor. If you agree with my approach, I would consider it ready to be merged. 

I will keep working on extracting more lexical data. Either in this PR or the next one.

src/wiktextract/datautils.py
```diff
@@ -6,6 +6,7 @@
 from collections import defaultdict
 from functools import lru_cache, partial
 from typing import Any, Dict, Iterable, List, Tuple
+from wiktextract.extractor.es.models import BaseModelWrap
```
xxyzz, line 9: This line would import `pydantic` and the Spanish extractor code for all other extractors code import this file, this is not ideal.

xxyzz: I made some changes in #401, maybe that's enough for removing boilerplate methods in `BaseModelWrap`.

src/wiktextract/extractor/es/models.py
empiriker: Thanks for the thorough review and pointing my in the right direction!

> The default value of extra is [ignore](https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.extra), maybe the base model should use **forbid.**

With the option **forbid**, we would get an exception every time there is an extra field. This would mean that we need to add error handling every time we instantiate a pydantic model. I think that would be extremely cumbersome and I didn't find a good way to catch this exception in an abstract way... There is an interesting PR suggesting a fourth option [**warn**](https://github.com/pydantic/pydantic/issues/4945) which would be ideal for our case.

> Isn't the JSON schema file generated from the defined model classes? How could they contain fields created dynamically at runtime?

You're right. The JSON schema file will only contain fields defined in the model classes.

> As for your German data example, I think we should not check the fields(or generate JSON schema) not created in our code.

The main question is thus: Do we want that our JSON schemas strictly describe our output, i.e. the output contains only keys that are described in our JSON schemas? Or do we not mind if the output contains extra fields?

My preference and also the PR's is the first option.

src/wiktextract/datautils.py
empiriker: That was a great change to remove the need for including the pydantic dependency everywhere.  :+1:

Regarding the boilerplate:
- The `.get()` method is needed as long as we use this in our utility functions such as in [append_base_data](https://github.com/tatuylonen/wiktextract/blob/1ccc547567b872df06019a372d006bc5519d1746/src/wiktextract/datautils.py#L249C8-L249C17) 
- For `__getitem__()`, the same applies, e.g. [here](https://github.com/tatuylonen/wiktextract/blob/1ccc547567b872df06019a372d006bc5519d1746/src/wiktextract/datautils.py#L250).
- The same for `__setitem__()`: [here](https://github.com/tatuylonen/wiktextract/blob/1ccc547567b872df06019a372d006bc5519d1746/src/wiktextract/datautils.py#L258)

- The `update()` function isn't strictly necessary but still useful to have. 

I will updated at least the new code to avoid dict like assignments.

src/wiktextract/extractor/es/models.py
```diff
@@ -0,0 +1,123 @@
...
+        try:
+            setattr(self, item, value)
+        except ValidationError:
+            pass
+
+
+class LoggingExtraFieldsModel(BaseModelWrap):
+    @model_validator(mode="before")
+    def log_extra_fields(cls, values):
+        all_allowed_field_names = {key for key in cls.__fields__.keys()}
```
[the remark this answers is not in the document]
empiriker: Thanks again! It's [`model_fields`](https://github.com/pydantic/pydantic/blob/b430e5aa9d1061e65753301ee014278eab66c437/pydantic/main.py#L102) that we need here. No idea, why they don't mention this property in the doc but it's clearly the replacement for `__fields__`.

> Maybe you'd like to make this debug log disabled by default, because there will be lots of debug messages spamming the log.

Interestingly, there shouldn't be a lot of spam from this logging call. At least how I see it used during development and maintaining of the extractors:
- If we set a field manually that isn't defined in a model, that's clearly an oversight and the model should be updated. To not miss this, the log can be useful and would disappear afterwards.
```
for key, value in template_node.template_parameters.items():
    if isinstance(key, str):
        reference_data[key.lower()] = clean_node(wxr, {}, value)
```
- In a case like above, the code suggests that the decision has been made to indeed capture all the fields. Than it's just a matter of running the extractor code once and update the models accordingly. Again, the debug messages would disappear afterwards. 
- If only same keys from that `template_node` are supposed to be extracted, then this should be done explicitly in code. Again not that many/no debug messages.

src/wiktextract/datautils.py
xxyzz: The new code using pydantic models shouldn't use `append_base_data()`, you could append the new `WordEntry` directly.

src/wiktextract/extractor/es/models.py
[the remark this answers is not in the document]
xxyzz: Undefined field shouldn't be allowed to be assigned. In your template parameters case, we could defined that field in its parent model as dictionary or `Any`.

src/wiktextract/datautils.py
empiriker: Thanks! I updated the code accordingly and indeed could get rid of the boilerplate.

Perhaps the `append_base_data()` function should then be moved to the chinese extractor code?

src/wiktextract/extractor/es/models.py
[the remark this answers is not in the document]
empiriker: I would like to avoid not specifying the keys in a dictionary. IMO, the JSON schema is there to inform users what to expect. Extracting data into an unspecified dict is perhaps still better than not extracting but it's immensely yes useful to many applications without knowledge what's in the dict.

I feel like we are still not on the same page. Probably due to a lack of good examples. Perhaps looking at my WIP branch `es-ahead` will help:

- [Here](https://github.com/empiriker/wiktextract/blob/ecf290e1c25ee630627793a3c9daf92c4b30ccde/src/wiktextract/extractor/es/example.py#L49-L68) I apply the template parameters case to parse a reference template. When I encounter an unknown key, I get a debug message and can update the [pydantic model](https://github.com/empiriker/wiktextract/blob/ecf290e1c25ee630627793a3c9daf92c4b30ccde/src/wiktextract/extractor/es/models.py#L92-L111).

Obviously, there are other ways to achieve the same effect. I thought it intuitive to also delegate the task of notifying us about the ignored fields to pydantic, since (with the option `extra="ignore`) it checks for unspecified fields anyway. But if you think it's a bad idea, I can get rid of `LoggingExtraFieldsModel`.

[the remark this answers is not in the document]
xxyzz: I didn't know you would break the template arguments that much... I don't know how pydantic model could fit in this case, maybe it shouldn't be used for dynamically created data because that code will break every time when the template is changed. I'm not sure... You could experiment with the code and see what is best for your usage cases.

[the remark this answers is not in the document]
xxyzz: And I don't understand how your code works... set `extra` to `ignore` only let pydantic ignore the validation error when creating the model instance, that extra data or later assigned data is still not added.

```python
from pydantic import BaseModel, ConfigDict, model_validator


class MyBaseModel(BaseModel):
    model_config = ConfigDict(validate_assignment=True)
    username: str


m = MyBaseModel(username="a", password="b")  # no password
getattr(m, "password")  # error
setattr(m, "foo", "bar")  # error
print(getattr(m, "foo"))  # error
```

Am I missed something?

src/wiktextract/extractor/es/models.py
xxyzz: Sorry I missed this comment. If the code is initializing a pydantic model with undefined fields and get exception, that means the code or the model is wrong and that's the whole point of validation, right? The only way we could get this exception is from our own mistake but not from the changes in Wiktionary.

src/wiktextract/extractor/es/models.py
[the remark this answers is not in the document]
empiriker: @xxyzz I am very grateful for your comments and that you're really thinking this through alongside with me but I have to admit that right now I am also a bit lost what your exact worries are. So I am hesitant to change something in the code right now without knowing exactly what I am addressing. 

Perhaps, I can try to address your points in writing and you could specify what needs to be addressed/changed in my proposal?

---
Just to make sure we are on the same page, here are the three different options [how a pydantic model deals with extra values](https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.extra):
- `allow` - Allow any extra attributes.
- `forbid` - Forbid any extra attributes.
- `ignore` - Ignore any extra attributes.

With the default value of `extra="ignore"`, your example does result in:

```
from pydantic import BaseModel, ConfigDict, model_validator


class MyBaseModel(BaseModel):
    model_config = ConfigDict(validate_assignment=True)
    username: str

m = MyBaseModel(username="a", password="b")  # (A) no error, password gets ignored
getattr(m, "password")  # (B) AttributeError: 'MyBaseModel' object has no attribute 'password'
setattr(m, "foo", "bar")  # (C) pydantic_core._pydantic_core.ValidationError: 1 validation error for MyBaseModel
foo
print(getattr(m, "foo"))  # (D)  AttributeError: 'MyBaseModel' object has no attribute 'foo'
```

(A) Works exactly as intended: 
- `password` gets ignored (we only get keys in our output that is described in our schema/classes)
- We do not get an error and the code just continues executing.
- When inheriting from `class LoggingExtraFieldsModel(BaseModelWrap)`, the extra field `password` would get logged and we could update the schema/class.
- And we still get type validation on the defined field `username`. 

(B) and (D) do not worry me:
- Instance attributes should be accessed as attributes: `m.username`
- That also makes for a better developer experience since my IDE tells me which attributes exist.
- We could, of course, overwrite the `getattr` function to return None for undefined keys (like `defaultdict`) but I don't see a point in this.
- If absolutely necessary, one could always use `try/except`.

(C) is in my eyes a bit problematic but solvable:
- Indeed, I realize now that my previous example of [`add_template_params_to_reference()`](https://github.com/empiriker/wiktextract/blob/ecf290e1c25ee630627793a3c9daf92c4b30ccde/src/wiktextract/extractor/es/example.py#L49-L68) does only work since I defined the dict-like setter using an exception:
```
class BaseModelWrap(BaseModel):
    def __setitem__(self, item, value):
        try:
            setattr(self, item, value)
        except ValidationError:
            pass
```
- Without redefining `__setitem__`, I would need to use `try/catch` locally.
- This would then mimic the behaviour of extra arguments on class instantiation: Extra arguments would just be ignored but logged, and an update of an existing key with the wrong type would simply not be executed. IMO, the behaviour we want. 

I hope that clear up the technical question marks.

---

I will now try to answer your concerns.

>  I don't know how pydantic model could fit in this case, maybe it shouldn't be used for dynamically created data because that code will break every time when the template is changed.

No. As I see it, nothing would break since unknown keys don't throw errors (case (A)) or would need to be caught anway (case (C)). If a template would change and introduce a new key, e.g. `{template|new=foo}`, we would simply not extract the `new` key and potentially get a debug message.

> Maybe you could only define ref field in model and copy template arguments to a dictionary field which won't be checked. 

I prefer making the structure explicit but, yes, that is a valid option that we can flexibly choose. For example, if you were to adopt pydantic for the French extractor you could type `ref` as an [iterable type](https://docs.pydantic.dev/1.10/usage/types/#typing-iterables):
```
class Example(BaseClass):
    ref: dict = None
```
or 
```
class Example(BaseClass):
    ref: Dict[str,str] = None
```
Or if you wanted to define some keys but not all you could overwrite the "extra" configuration:
```
class Reference(BaseClass):
    model_config = ConfigDict(extra="allow")
    title: str
    
class Example(BaseClass):
    ref: Reference
    
ref = Reference(title="Hamlet",author="Shakespeare")
print(ref.model_dump()) 
# {"title":"Hamlet", "author":"Shakespeare"}
```

> If you use pydantic this way, the code will be much longer and require more efforts to maintain, and therefore it's worse then our current defaultdict approach.

I fail to see it. It might a little bit more work initially but, IMO, is much more maintainable than manual JSON schemas. 

Once some base classes are defined for the Spanish extractor, they can also be reused for other extractors via inheritance while still easily specifying where they differ, e.g.:
```
class PronunciationBase(BaseClass):
    ipa: str
    
class SpanishPronunciation(PronunciationBase):
    syllabic: List[str] = Field(default=[], description="Syllabic transcription")

class RussianPronunciation(PronunciationBase):
    stress: str = Field(None, description="Word with stress markings")
```

Or am I missing something?

> [dynamic model creation](https://docs.pydantic.dev/latest/concepts/models/#dynamic-model-creation)
I wasn't aware of that. Thanks! It's a cool feature and we could theoretically use it to output a dynamic schema at the end of each extraction run. But it would be a hassle with multithreading and add very little benefit in the end...

> And we already have a [JSON data structure browser](https://kaikki.org/dictionary/errors/mapping/index.html) page on kaikki.org.

Very useful indeed! But as I understand it, this is more exploratory in its nature? It still doesn't replace a proper and detailed JSON schema. 

Perhaps lastly, let me explain my use case: I want to convert the JSON Schema for each extractor to a Typescript type (either from the schema or from pydantic directly ([`pydantic-to-typescript`](https://pypi.org/project/pydantic-to-typescript/))) and then use these types to build a frontend, flexibly displaying lexical information extracted from wiktextract. Having these types during development will make things so much easier. 

This is why I think our schema should be as detailed as possible. I am almost certain, I wouldn't use a field like `ref: Dict[str,str] = None` since I had no idea what's in there. But in 
```
class Reference(BaseClass):
    title: str
```
I might at least use the `title` field.

---
I hope that cleared things up. 

Where do see blocking issues at the moment?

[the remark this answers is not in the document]
xxyzz: Our main disagreement is our priority. Our priority is getting more data from the dump file, but your priority is getting JSON schema. Your code can't adapt to changes on Wiktionary, it needs constant maintenance to sync models with templates to add new template arguments. Because you're using pydantic model on data that shouldn't be validated.

You can get better JSON schema by running through the generated jsonl file like what kaikki.org did. And if you're trying to use dynamic model creation, that'd be just like scanning the jsonl file, but more meaningless validation  and slower.

Back to my example code, all 4 errors are bugs in our Python code and shouldn't be ignored. Set `forbid`, we get error log with one line code. You have to loose the validation on data you don't have control but have to add and harden the validation on data you do control. I think you're pursuing a goal that pydantic can't fulfill. But I haven't use pydantic before, I could be wrong.

I think you have a minor misunderstanding about the `ref` field in other extractors, it is a `str`, the plain text of the example sentence source. In the French extractor there are no template arguments dictionary but the English extractor has many. `dict[str, str]` or `Any` is exactly what I suggested before but you choose to use a pydantic model. It's even more easier if you don't translate the arguments to English but just copy them.

We could get this pr merge if you could update the typing, some are still using `Dict`. I think `process_pos_block()` could get `index out of range` error because `page_data` is empty.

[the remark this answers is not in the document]
empiriker: Okay. Got it. 

I think there is a way to have our cake and eat it when it comes to template arguments dictionaries. But let's deal with it when I get to it. If it won't work out then, I am fine with using `dict[str,str]` for these kind of fields.

> I think you have a minor misunderstanding about the ref field in other extractors, it is a str, the plain text of the example sentence source. 

I am aware of this. At the very beginning when I started contributing here, someone told me that you do not want to enforce a strict schema for all extractors but rather see what makes sense for each Wiktionary edition. So while modelling the new extractors on the existing ones, I didn't think it would hurt to be more specific. In the case of `ref`, why only extract the plain text if there is more structured data to have? In the[ German extractor](https://github.com/tatuylonen/wiktextract/blob/295f9100ee05ac644492a490d9c8c0f6177720b5/src/wiktextract/extractor/de/example.py#L53-L80), for example, I have stored the plain text with the `raw_ref` within a `ref` dictionary.

> We could get this pr merged if you could update the typing, some are still using Dict. I think process_pos_block() could get index out of range error because page_data is empty.

Done. `process_pos_block` doesn't get an index out of range error since it appends a copy of `base_data`  into `page_data` before using index access.

xxyzz:
Thanks for your contribution!

kristian-clausal:
Thanks for your contribution from me too!

PR #397, opened 2023-11-13
Add Containerfile

xxyzz:
Run code in a container could solve all the environment errors. We could publish the container image to [GitHub Packages](https://github.com/tatuylonen/wiktextract/packages) so the users don't need to build the image. This could also relief the burden of compatible with older Python verisons.

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -1,0 +1,22 @@
+name: Publish container
+
+on:
+  workflow_run:
+    workflows: [test]
+    types:
+      - completed
+    branches:
+      - master
+
+jobs:
+  publish_container:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - uses: actions/checkout@v4
+      - run: podman login -u ${{ github.repository_owner }} -p ${{ secrets.GITHUB_TOKEN }} ghcr.io
+      - run: podman build -t ghcr.io/${{ github.repository }} .
+      - run: podman push ghcr.io/${{ github.repository }}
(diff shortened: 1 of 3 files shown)


xxyzz:
The command of installing package from pypi.org is deleted because that package is so outdated.

kristian-clausal:
This seems good for those users who wish to use containers, I don't see why we couldn't have this. Is the image generation possible to automate so that upkeep of a GitHub package is hassle-free?

xxyzz:
Maybe just need to run two `podman` commands to build and upload the container image. I'll try the commands in my fork repo.

kristian-clausal:
Thanks, this seems great.

xxyzz:
The published container image page in my fork repo: https://github.com/xxyzz/wiktextract/pkgs/container/wiktextract

empiriker:
I always appreciate containerized applications when I just want to run an application as is. So this is a good addition. :+1: 

However, I don't think it solves all issues with versioning, in general. In the past, I have used wiktextract as a dependency and then overwritten or customized parts of the application and I always struggled with using a specific version of wiktextact. 

In particular, I used to fix the version of wiktextract by specifying a certain commit, e.g.: `wiktextract @ git+https://github.com/tatuylonen/wiktextract.git@a3665b8779ad78b045406cc7b1e9ce31876bc6b6`. 

The problem is that the install script of `wiktextract` always installs the latest version of `wikitextprocessor` from the repository. After a while a fresh install will result in incompatibile versions of `wikitextprocessor` and the commit of `wiktextract`. 

As I understand, the containerized application won't have that problem. But in the described use case, where the wiktextract isn't run as is but used as a dependency, the container won't help.

IMO, the ideal solution would be to introduce proper versioning for wiktextract's dependecies.

xxyzz:
You could use `pip freeze` and install the packages with the `--no-deps` option.

I don't think we could make releases for both packages, because Wiktionary changes every day and our parser is unstable.

PR #410, opened 2023-12-01
Extract pronunciation from Spanish Wiktionary

empiriker:
This is basically to parse the [`{pron-graf}`](https://es.wiktionary.org/wiki/Plantilla:pron-graf) template and extract pronunciation data. The template itself is very vast and covers also different spellings.

I am not 100% whether extracting spelling data to a separate `spelling` field is the best way to proceed. But it's a start.

### DIFF
--- a/json_schema/es.json
+++ b/json_schema/es.json
@@ -56,6 +56,143 @@
       ],
       "title": "Sense",
       "type": "object"
+    },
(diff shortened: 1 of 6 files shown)


kristian-clausal:
Thank you for contribution!

PR #411, opened 2023-12-03
Parse page for Russian Wiktionary

empiriker:
Hi there!

I have started to work on extracting the Russian Wiktionary and would like to hear your advice. It's really fascinating how each Wiktionary edition comes with its own set of intricacies and difficulties. At first glance, the Russian Wiktionary looked very structured and promised to be an easy job. At second glance, the structure can be quite complicated and differs from what we worked with so far. 

The two issues I butted my head with and would like to hear your input are: Homonyms/homographs and POS. 

# Homonyms/Homographs
The Russian Wiktionary deals quite differently with homonyms and homographs than the other editions. Let's look at the [page structure](https://ru.wiktionary.org/wiki/%D0%92%D0%B8%D0%BA%D0%B8%D1%81%D0%BB%D0%BE%D0%B2%D0%B0%D1%80%D1%8C:%D0%9F%D1%80%D0%B0%D0%B2%D0%B8%D0%BB%D0%B0_%D0%BE%D1%84%D0%BE%D1%80%D0%BC%D0%BB%D0%B5%D0%BD%D0%B8%D1%8F_%D1%81%D1%82%D0%B0%D1%82%D0%B5%D0%B9) first:

- H1 headings are used to separate different language sections for the same lemma
- H2 headings are optional and used if the same lemma can refer to different homonyms/homographs
- H3 headings are then used to structure each entry: Syntactic properties, semantic properties, pronunciation etc.

The H1 and H3 headings are great and similar to what we can find in other Wiktionary editions. The problem is with the H2 headings.

To my knowledge, the other editions do disambiguate homonyms and homographs only where they have different parts of speech. The Russian Wiktionary is more granular. It uses the [`{заголовок}`](https://ru.wiktionary.org/wiki/%D0%A8%D0%B0%D0%B1%D0%BB%D0%BE%D0%BD:%D0%B7) template to generate the H2 headings which distinguishes different POS and stress patterns. 

I think the three examples from the template illustrate my dilemma quite well:

```
омонимы: бита
== {{з|(существительное I)}} == # noun I
== {{з|(существительное II)}} == # noun II
== {{з|(существительное III)}} == # noun III
== {{з|(существительное IV)}} == # noun IV
== {{з|(существительное V)}} == # noun V
== {{з|(причастие)}} == # participle
омографы: гонит
== {{з|ударение=гони́т}} == # stress=гони́т
== {{з|ударение=го́нит}} == # stress=го́нит
омонимы + омограф: коса
== {{з|ударение=коса́ (существительное I)}} == # stress=коса́, noun I
== {{з|ударение=коса́ (существительное II)}} == # stress=коса́, noun II
== {{з|ударение=коса́ (существительное III)}} == # stress=коса́, noun III
== {{з|ударение=коса́ (прилагательное)}} == # stress=коса́, adjective
== {{з|ударение=ко́са|I}} == # stress=ко́са
== {{з|ударение=ко́са|II}} == # stress=ко́са
```

In other editions, the first example would just consist of two section бита (noun) and бита (participle) and noun I to noun V would be senses of the main (noun) entry. However, in the Russian edition each noun I to noun V tends to have all the lexical information a page entry usually has; what distinguishes the Russian noun I-V sections from a usual sense section is for example that they have each a different etymology and can each have multiple glosses (senses).

The second example is similar in that the different stresses would just be part of the same lemma+POS entry.

The third example is interesting in that it shows the different combinations and that this template doesn't necessary specify the POS (!).

So my question is: Should I attempt to map this to our usual page_data structure or keep the separation of the Russian Wiktionary?

To exemplify, should the `page_data array` of the first example be:

`[{"word":"бита", "pos":"noun", "senses":[sense1, ..., sense5]}, {"word":"бита", "pos":"participle", "senses":[sense1]}]` where each `senseN` would have it's own etymology and different subsenses

or

`[{"word":"бита", "pos":"noun", "senses":[senses...]}, [{"word":"бита", "pos":"noun", "senses":[senses...]}, [{"word":"бита", "pos":"noun", "senses":[senses...]}, [{"word":"бита", "pos":"noun", "senses":[senses...]}, [{"word":"бита", "pos":"noun", "senses":[senses...]},  {"word":"бита", "pos":"participle", "senses":[senses]}]`  where `senses` are just "simple" senses of that particular homonym/homograph?

# Part of speeches

The second difficulty I had is the way the Russian Wiktionary specifies the POS. It does so in two places: the [`{заголовок}`](https://ru.wiktionary.org/wiki/%D0%A8%D0%B0%D0%B1%D0%BB%D0%BE%D0%BD:%D0%B7) template and the H3 section "Морфологические и синтаксические свойства"/"Morphological and syntactic properties".

Since the first one only exists if a page's lemma has homonyms/homographs and even then the POS is not necessarily specified, I looked to extract the POS from the *Morphological and syntactic properties* section. This indeed seems to be the universal (at least in intent) place where POS are given.

However, in most cases the POS is generated from a from template of which there are tens of thousands (different template names). For example, there are 3780 templates that start with `{сущ ...}` which (according to some random observation) all generate a forms table and somewhere a string "существительное" (noun) with further specifications of the type of noun. But instead of having different arguments to a common `сущ` template the arguments are part of the template name. 

So what I have to do is search the template name for different substrings that might indicate a POS. This is what `POS_MAP` is for which I created iteratively. The problem is that this is not very precise and in theory one of these "pos substrings" could appear in a template's name which has a completely different meaning. 

To further complicate things, usage of a template is not mandatory (though by far the general case). Sometimes a POS is specified in a link, simply in plain text, or perhaps a template that wasn't captured by `POS_MAP`. So I resorted to the backup plan of just cleaning the "Morphological and syntactic properties" section and searching for POS names in the text. Again not ideal for obvious reasons.

My questions:
- Do you have any ideas what else I could try that might generalize better?
- If it turns out that we don't find a better method than what I currently do: What should we do in case there is no POS found?

---

Thanks a lot for your input!


empiriker:
Thanks for your review and feedback!

In this case, this seems to be best we can do for now and it should get the job done. 

At the risk of jinxing it: I think extracting the individual sections should be straightforward now (at least the default case). 

> I'd suggest only match the first text in the split template names since all keys in POS_MAP seem to be at the start of the template name.

There is a small minority of template names where the part specifying the POS is in the second position. I restricted the loop to only look at the first two splits max.

xxyzz:
Could you please set the `extra` to `forbid` in the base model so `PydanticLogger` and `LoggingExtraFieldsModel` could be removed?

empiriker:
Done. At this point, I am fine with being more restrictive and removing the logging. 

This [code snippet](https://github.com/empiriker/wiktextract/blob/47f57bba9ed7f58d8daf75649ce3ed6e29eb4740/src/wiktextract/extractor/es/example.py#L63-L70) from the current Spanish PR #414 convinced me that it is indeed unnecessary to delegate the task of tracking undefined keys to pydantic. It should be done in code instead.

xxyzz:

kristian-clausal:
Thank you!

PR #413, opened 2023-12-04
Move code out of for loop that didn't belong there

kristian-clausal:
Fixes an issue where the index of `audios` was out of range because we emptied it in an earlier iteration of the loop. (Line 566 with the IPA audio file stuff, `audios` is emptied on line 639).

This was my fault from last year, for some reason this bit of the code that should be handled just once was put inside a large for loop, and some kind of brainfart must have made me think that "this should run only once, let's prevent the if-block from happening by emptying `audios` after we've done the processing"... And I didn't think that `audios` was accessed elsewhere in the loop, which triggered this pretty rare bug.

I'm putting this here overnight just in case I missed something obvious. But I'm pretty sure this is just my own mistake originally...

### DIFF
--- a/src/wiktextract/pronunciations.py
+++ b/src/wiktextract/pronunciations.py
@@ -424,7 +424,7 @@ def parse_pronunciation(wxr, node, data, etym_data,
     active_pos = None
 
     for text, ipa_text in split_cleaned_node_on_newlines(contents):
-        # print(text, ipa_text)
+        # print(f"{text=}, {ipa_text=}")
         prefix = None
         if not text:
             continue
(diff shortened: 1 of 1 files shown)


src/wiktextract/pronunciations.py
```diff
@@ -582,72 +582,72 @@ def split_cleaned_node_on_newlines(contents):
         # XXX what about {{hyphenation|...}}, {{hyph|...}}
         # and those used to be stored under "hyphenation"
 
-        # Add data that was collected in template_fn
-        if audios:
```
xxyzz, line 586: `if` could be removed, the `for` loop won't run when `audios` is empty. The problem of this code is this function is too long and has too many nested functions. It's hard not to make indent mistakes or override shared variables in code like this.

PR #416, opened 2023-12-05
Add a script to create all JSON schema files from pydantic models

xxyzz:


empiriker:
That's great!

Is there a way to run this script automatically with a GitHub action and commit the latest JSON schemas? I have to admit, it has been more than once that I forgot to update the JSON schemas after introducing some changes.

empiriker:
Depending on the order of merging the outstanding PRs, in particular #411 , you might need to also remove the script from `ru/models.py`.

If you decide to merge this PR #416 first, I can remove it from there instead.

xxyzz:
We could use actions, I'll take a look, it's just really annoying to test action jobs. I'll merge your pr first and remove the code in this pr.

PR #417, opened 2023-12-05
Extract translations from Spanish Wiktionary

empiriker:
Perhaps worth a discussion: I am reusing an utility that I wrote for the German Wiktionary: `split_senseids()`. This function currently is defined in `extractor/de/utils.py`. 

It would make sense to move it somewhere else since it's not  used exclusively by the German extractor anymore. Any suggestions?

### DIFF
--- a/json_schema/es.json
+++ b/json_schema/es.json
@@ -1,6 +1,197 @@
 {
   "$defs": {
+    "Example": {
+      "additionalProperties": false,
+      "properties": {
+        "ref": {
+          "anyOf": [
+            {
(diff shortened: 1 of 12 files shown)


kristian-clausal:
These kinds of shared bits could (and maybe should) be put into a shared utility module, yeah. They might need a unique name (split_senseids1, split_senseids2...) in case there are going to be more similar kinds of functions with an obvious similar name, and then they could be imported with "from x import y1 as y".

It could be extremely messy, too. At the moment that is definitely not a problem, but in the future...

PR #418, opened 2023-12-05
Add pydantic and json_schema to German extractor

empiriker:
This is to add pydantic classes to the German extractor.

In some cases, pydantic fields are defined as lists of strings where probably they could be just a string. I decided to accept for now this looser schema to avoid refactoring even more. This was already quite a bit more painful than I anticipated.

### DIFF
--- a/json_schema/de.json
+++ b/json_schema/de.json
@@ -1,0 +1,932 @@
+{
+  "$defs": {
+    "Example": {
+      "additionalProperties": false,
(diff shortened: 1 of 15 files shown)


xxyzz:
Thanks for your contribution!

How's the performance compare to the previous code?

xxyzz:
That's pretty insignificant. We probably don't need to worry about it.

PR #419, opened 2023-12-05
Extract pronunciation from Russian Wiktionary

empiriker:
More or less straightforward. I skipped some of the rarer pronunciation templates but indicated them in code with `# XXX:`

PS: I hope you don't mind the many simultaneous PRs. This is the last week where I can really focus on contributing to wiktextract, so I try to make the most of it.

### DIFF
--- a/json_schema/ru.json
+++ b/json_schema/ru.json
@@ -1,7 +1,135 @@
 {
+  "$defs": {
+    "Sound": {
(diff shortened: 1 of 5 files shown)


xxyzz:
We are happy to see more contributions. And thank you for implementing these new extractors and adding pydantic models.

PR #420, opened 2023-12-06
Extract linkages from Spanish Wiktionary

empiriker:
This is as far as I planned to develop the Spanish extractor. 

Apart from places indicated with `# XXX: ...`, the biggest area of improvement would be perhaps to deal with form sections. In the Spanish Wiktionary, whenever a lemma can also be a conjugated/declinated form of a different word, this will be noted in a form section which are siblings to POS sections.

To get an order of magnitude, in the 20,000 pages I am developing on, I reach these form sections ca. 15,000 times. An example, for the word [pueblo](https://es.wiktionary.org/wiki/pueblo):

image

For me it's okay to not deal with these right now. But perhaps someone else will.

### DIFF
--- a/json_schema/es.json
+++ b/json_schema/es.json
@@ -40,6 +40,45 @@
       "title": "Example",
       "type": "object"
     },
+    "Linkage": {
+      "additionalProperties": false,
+      "properties": {
+        "alternative_spelling": {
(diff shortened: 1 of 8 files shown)


xxyzz:

PR #421, opened 2023-12-06
Add JSON schema files to GitHub Pages

xxyzz:
- Replace nose2 with unittest
- Ignore the `taxondata.py` file
- Upload JSON schema files generated from pydantic models to GitHub Pages


empiriker:
That looks like a good automation. :+1: 

Do I understand it correctly that the schema files wouldn't be part of the repo anymore but need to be downloaded manually from this repo's GitHub pages? Then it will be a bit unfortunate that not everything is in one place anymore. But I guess that's a price worth paying for always current schemas.

Out of curiosity: Why the change from nose2 to unittest?

PR #422, opened 2023-12-06
Spanish etymology extraction

999pingGG:
As promised, here's my first contribution! I tried to follow already-established patterns and style. Please, let me know about any improvements I can make, as small as it could be, I like doing things the right way.

### DIFF
--- a/tests/test_es_etymology.py
+++ b/tests/test_es_etymology.py
@@ -1,0 +1,100 @@
+import unittest
+
+from wikitextprocessor import Wtp
+from wiktextract.config import WiktionaryConfig
+from wiktextract.extractor.es.etymology import process_etymology_block
+from wiktextract.extractor.es.models import WordEntry
+from wiktextract.wxr_context import WiktextractContext
+
+
+class TestESEtymology(unittest.TestCase):
+    def setUp(self) -> None:
+        self.wxr = WiktextractContext(
+            Wtp(lang_code="es"), WiktionaryConfig(dump_file_lang_code="es")
+        )
+        self.wxr.wtp.add_page(
+            "Plantilla:etimología",
+            10,
+            """{{#switch: {{{1|}}}
+ |=Si puedes, incorpórala: ver cómo.
+ |Del {{{1|}}} {{{2|}}}
+}}""",
+        )
+        self.wxr.wtp.add_page(
+            "Plantilla:etimología2",
+            10,
+            "{{{1|Si puedes, incorpórala: ver cómo.}}}",
+        )
+
+    def tearDown(self) -> None:
+        self.wxr.wtp.close_db_conn()
+
+    def test_es_extract_etymology(self):
+        test_cases = [
+            {
+                # https://es.wiktionary.org/wiki/Schreck
+                "input": "{{etimología|leng=de}}",
+                "expected": {
+                    "etymology_templates": [
+                        {
+                            "args": {
+                                "leng": "de",
+                            },
+                            "name": "etimología",
+                            "expansion": "Si puedes, incorpórala: ver cómo.",
+                        },
+                    ],
+                },
+            },
+            {
+                # https://es.wiktionary.org/wiki/bagre
+                "input": "{{etimología|catalán|bagre}}, y este del latín ''[[pargus]]'', a su vez del griego ''[[φάγρος]]'' (''phágros'')",
+                "expected": {
+                    "etymology_templates": [
+                        {
+                            "args": {
+                                "1": "catalán",
+                                "2": "bagre",
+                            },
+                            "expansion": "Del catalán bagre",
+                            "name": "etimología",
+                        },
+                    ],
+                    "etymology_text": "Del catalán bagre, y este del latín pargus, a su vez del griego φάγρος (phágros)",
+                },
+            },
+            {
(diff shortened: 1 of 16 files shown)


empiriker:
First of all, congrats and thanks for contributing!

Only a couple of things I noted right now. I will take a closer look tomorrow:
- Use `dict` instead of `Dict` for typing. We only recently dropped support of python 3.8. So in old code you will still see `Dict`, but new code should use `dict`> The same for `list`/`List`.
- I'm surprised, you're creating a new `WordEntry` instance. Usually it gets created once for each POS and then passed to all other extraction methods, so that they *add* to that instance. Probably that's what you want to do here.
- Consider moving `process_etymology_block` to a new file `etymology.py`
- Consider adding test cases for the most common scenario(s): it can help during development, makes code review easier, and makes everything more maintainable

999pingGG:
My test fails the `etymology_text` check, because it involves expanding a template. Expanding it during a real run is successful, but in the test it expands to nothing. I guess the `WiktextractContext` needs to know about the template(s), how can I do this?

tests/test_es_etymology.py
```diff
@@ -0,0 +1,98 @@
...
+                result_etymology_templates = result.get("etymology_templates")
+                expected_etymology_templates = case["expected"].get("etymology_templates")
+
+                self.assertTrue(len(result_etymology_templates) == len(expected_etymology_templates))
+                for i, expected_template in enumerate(expected_etymology_templates):
+                    result_template = result_etymology_templates[i]
+
+                    self.assertEquals(expected_template["name"], result_template["name"])
+                    if "args" in result_template:
+                        self.assertDictEqual(expected_template["args"], result_template["args"])
```
xxyzz, line 95:
```suggestion
                    self.assertEqual(expected_template["name"], result_template["name"])
                    if "args" in result_template:
                        self.assertEqual(expected_template["args"], result_template["args"])
```

`assertEquals` is removed in Python 3.12 and `assertEqual` calls `assertDictEqual` when passed dictionaries.

You could use [`Wtp.add_page()`](https://github.com/tatuylonen/wikitextprocessor/blob/619981b7f478d2d1479494ca69e1796865afe134/src/wikitextprocessor/core.py#L778-L786) to add template with expanded body in test:


```python
self.wxr.wtp.add_page("Template:name", 10, "expanded text")
```

empiriker:
Having taken another look at the page structure when there are multiple etymologies in a language section, I realized that the current code won't extract etymologies in a  `=== Etymology # ===` section. I initially thought these would be followed from another subheading `==== Etymology ====` but it rather seems that the etymology data comes directly after the H3 heading and before the next H4 heading.

https://github.com/tatuylonen/wiktextract/blob/c7094e70a930584c560dcfd3febd51901de06b48/src/wiktextract/extractor/es/page.py#L111-L112

tests/test_es_etymology.py
999pingGG: How about now? I couldn't get the template to actually expand its parameters so I hardcoded a generic expansion for the templates, like other tests do. Also, I found and fixed a flaw in the logic thanks to the test 😅

999pingGG:
@empiriker Yes, numbered etymology sections are currently ignored, but I hoped to contribute an initial implementation first, then discuss how those sections should be handled because it's unclear to me.

Looks like the Wiktionary entries can have more than one etymology section due to the same word having several origins and several meanings per origin, but you have to pick one. How do you decide which one? Why not extracting everything?

xxyzz:
I find a page has two Etimología sections: https://es.wiktionary.org/wiki/embazarse

Use the same structure as the wikitext is easier to implement. We could save each Etimología section in their own `WordEntry`.

tests/test_es_etymology.py
```diff
@@ -0,0 +1,97 @@
...
+                expected_etymology_templates = case["expected"].get("etymology_templates")
+
+                self.assertTrue(len(result_etymology_templates) == len(expected_etymology_templates))
+                for i, expected_template in enumerate(expected_etymology_templates):
+                    result_template = result_etymology_templates[i]
+
+                    self.assertEqual(expected_template["name"], result_template["name"])
+                    if "args" in result_template:
+                        self.assertDictEqual(expected_template["args"], result_template["args"])
+                    self.assertEqual(expected_template.get("expansion"), result_template.get("expansion"))
```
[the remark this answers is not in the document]
999pingGG: Ah, of course, manually checking members was a leftover from when I tried to check only some of them and ignore template expansion.

tests/test_es_etymology.py
```diff
@@ -0,0 +1,92 @@
...
+                process_etymology_block(self.wxr, page_data[0], root)
+
+                result = page_data[0].model_dump(exclude_defaults=True)
+
+                self.assertEqual(case["expected"].get("etymology_text"), result.get("etymology_text"))
+
+                result_etymology_templates = result.get("etymology_templates")
+                expected_etymology_templates = case["expected"].get("etymology_templates")
+
+                self.assertTrue(result_etymology_templates, expected_etymology_templates)
```
xxyzz, line 91:
```suggestion
                self.assertEqual(result_etymology_templates, expected_etymology_templates)
```

xxyzz:

kristian-clausal:
Thank you!

PR #427, opened 2023-12-12
Improve French extractor 

xxyzz:


src/wiktextract/extractor/fr/page.py
```diff
@@ -99,7 +99,11 @@ def parse_section(
             ):
                 pass
             elif section_type in wxr.config.OTHER_SUBTITLES["notes"]:
+                if len(page_data) == 0:
+                    page_data.append(copy.deepcopy(base_data))
                 extract_note(wxr, page_data, level_node)
+                if page_data[-1].keys() == base_data.keys():
```
kristian-clausal, line 105: I didn't realize Python keys() returned a view object and that view objects were "set-like". I wonder if I could get Tatu to buy the library a new Python 3 reference...

I also found out you can do these kinds of comparisons with .keys() and .items(), but not .values() (unless the underlying object is exactly the same). See me forget this fact and try to do it some day...

xxyzz: I don't know that either, but I should migrate the French and Chinese extractor to pydantic and this code will be changed.

You can't compare `dict.values()`, it always return `False`: https://docs.python.org/3/library/stdtypes.html#dict.values

PR #432, opened 2023-12-21
Wiktextract typing

kristian-clausal:
I'm going to start slowly typing Wiktextract too (I already found some stuff that I maybe should have PRed separately, but it's such a small issue it doesn't matter at all: --statistics isn't actually implemented).

Keep on trucking with whatever is being done, I will manually merge if need be.

This is low priority, but really nice to have.

Current workflow: alphabetically. categories.py, config.py...


src/wiktextract/config.py
```diff
@@ -145,13 +173,14 @@ def set_attr_from_json(
...
         self.set_attr_from_json("POS_SUBTITLES", "pos_subtitles.json")
-        self.POS_TYPES = set(x["pos"] for x in self.POS_SUBTITLES.values())
-        for k, v in self.POS_SUBTITLES.items():
-            if "tags" in v:
-                assert isinstance(v["tags"], (list, tuple))
+        if self.POS_SUBTITLES is not None:
+            self.POS_TYPES = set(x["pos"] for x in self.POS_SUBTITLES.values())
+            for k, v in self.POS_SUBTITLES.items():
+                if "tags" in v:
+                    assert isinstance(v["tags"], (list, tuple))
```
xxyzz, line 180: `set_attr_from_json` set the attribute to empty dictionary if the JSON file doesn't exist, so there we don't need to check `POS_SUBTITLES` is `None`. And the `assert` is also unnecessary because `json.load()` can't create `tuple`.

src/wiktextract/config.py
```diff
@@ -96,32 +114,42 @@ def __init__(
...
+        self.debugs: list[ErrorMessageData] = []
+        self.redirects: SoundFileRedirects = {}
         self.data_folder = files("wiktextract") / "data" / dump_file_lang_code
+        self.POS_SUBTITLES: Optional[dict[str, POSSubtitleData]] = None
+        self.POS_TYPES: Optional[set[str]] = None
+        self.LINKAGE_SUBTITLES: Optional[dict[str, str]] = None
+        self.OTHER_SUBTITLES: Optional[dict[str, Union[str, list[str]]]] = None
+        # set the above three in the function below
         self.init_subtitles()
+        self.ZH_PRON_TAGS: Optional[list[str]] = None
```
xxyzz, line 134: All these variables are never be `None` actually. If they have to be defined here, use empty list/dictionary might be more appropriate.

src/wiktextract/clean.py
```diff
@@ -1423,9 +1439,12 @@ def repl_1_syntaxhighlight(m):
     title = re.sub(r"\[//[^]\s]+\s+edit\s*\]", "", title)
     # Replace links by their text
 
-    category_ns_data = wxr.wtp.NAMESPACE_DATA.get("Category", {})
-    category_ns_names = {category_ns_data.get("name")} | set(
-        category_ns_data.get("aliases")
+    category_ns_data: Optional[NamespaceDataEntry]
+    # XXX "Category" -> config variable for portability
+    category_ns_data = wxr.wtp.NAMESPACE_DATA.get("Category", None)
```
xxyzz, line 1444: `dict.get()` returns `None` by default if a default value is not passed.

kristian-clausal: I guess that is true, but it feels naughty.

src/wiktextract/config.py
kristian-clausal: I'll try using pre-declaration (`self.POS_SUBTITLES: dict[str, POSSubtitleData]\n`) and assertions. Should be fine in __init__, this object is created once.

src/wiktextract/type_utils.py
```diff
@@ -0,0 +1,14 @@
+from typing import (
+    Union,
+)
+
+
+WordData = dict[str, Union[
```
xxyzz, line 6: I think `list[dict[str, any]]` is fine for `parse_page()`, since these functions are not directly called and the data are written to JSON file soon after. Plus mypy doesn't work well with union types.

And could you please format the newly added code and rebase the commits instead of merge master branch?

kristian-clausal: What's the difference between rebasing and merging here?

xxyzz: rebase is like the branch is created at the current master branch so the commits will looks like they are created linearly otherwise old commits could appear anywhere after merge and it's harder to read the git log history, and rebase don't have distracting merge commits. Git book has some nice pictures: https://git-scm.com/book/en/v2/Git-Branching-Rebasing 

Basically I think it's like changing your start point to another branch, like the current master branch but not the outdated branch. And also has other features like squash commits and reword a commit message.

kristian-clausal: Tried rebasing, take a look at the commit history now. :+1: I will soon undo this and revert to previous commits, but suffice to say I will not be using rebase for this branch. I might restart the branch just to keep things simple and try, *try* if rebasing makes any sense then.

xxyzz: That's much better now, however there are still two extra merge commits. You could create a new branch from the `wiktextract-typing` branch locally and experimenting on that branch, then delete the old  `wiktextract-typing` and rename the new branch.

A comment to https://github.com/tatuylonen/wiktextract/pull/432/commits/0d41c02efb894a1d77265bbb71238e93f2016608: the long  `HEAD_TAG_RE` was created in a function before is because language data are loaded later from JSON files so they are not available from the beginning.

kristian-clausal: I've reverted to before rebasing, the duplicate commits are now gone.

xxyzz: I mean "were". Language data only come from the new package now.

My sentences are pron to have typos and grammatical errors if I'm not careful...

kristian-clausal: Phew, I thought I might have missed something.

xxyzz: I don't understand how could you merge these commits and start over?

kristian-clausal: I've merged the branch and now I'll start a new branch from master. Then I can use rebase to keep the branch updated without weird duplicate commits that happened when I rebased previously when I already had merges there. When I rebased previously, there were suddenly 18 commits instead of the actual 12, with duplicates. I would probably have to undo all the merges somehow to make a 'clean' branch and rebase that.

Continuing making small type checking changes in a new branch should be fine. It's not like it's actually code that breaks things.

xxyzz: I didn't think of the merge commits, not sure how to rebase in that case...

empiriker: You could start a new branch from the remote master branch and then cherry-pick individual commits from your working branch. 

It's a bit of work and a good lesson for the future to keep the git history clean. 😁 A lesson I had to learn myself recently. 😇

PR #433, opened 2023-12-22
Don't remove bold and italic wikitext before expanding nodes

xxyzz:
Remove the bold wikitext causes error in English Wiktionary template "ja-usex". Bold and italic nodes are converted to plain text later somewhere in `clean_value()`.

Resolves tatuylonen/wikitextprocessor#170


kristian-clausal:
Looks good to me.

PR #444, opened 2024-01-02
Remove pos length check in `test_long`

xxyzz:
Some pages use the template `zh-see` to redirect simplified Chinese characters don't have POS title. Not sure why the pr 174 of wikitextprocessor affects this test.

Another change is adding the redirect data to `pos_data` if it's not empty.


kristian-clausal:
Let's merge this if it fixes the issue with the test first, then wonder about errors later.

PR #445, opened 2024-01-03
Add daily schedule test action

xxyzz:
Trigger test action if wikitextprocessor is updated after the last test.


xxyzz:
The test action could also be triggered manually.

PR #446, opened 2024-01-03
 Typing stuff

kristian-clausal:
Typing stuff, part two, this time so it's possible to rebase (memo to self: don't merge to update, rebase this branch as you go).

I already found a bug. Is there anything worse than looking at some code, seeing that it's wrong, and then finding out it was your own commit that did it?

### DIFF
--- a/src/wiktextract/clean.py
+++ b/src/wiktextract/clean.py
@@ -9,13 +9,12 @@
 import re
 import html
 import unicodedata
-from typing import (
-    Callable,
-    Optional,
-    Union
-)
+from typing import Callable, Optional, Union
 from wikitextprocessor.common import MAGIC_FIRST, MAGIC_LAST
-from wikitextprocessor.core import NamespaceDataEntry
+from wikitextprocessor.core import (
+    NamespaceDataEntry,
+    TemplateArgs,
+)
 from .wxr_context import WiktextractContext
 
 ######################################################################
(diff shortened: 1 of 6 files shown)


kristian-clausal:
The point of type-checking all of this is to type-check things, and it feels silly to have to add `# type:ignore` for the outgoing data. We "know" what the data should be like, so we should be able to type-set it.

I will *attempt* to do some typing for the data. It *should* be possible, using TypedDict. TypedDict can (in 3.9) have optional and required fields (which are created by basically inheriting). We'll see if I can make it work out, or if I can just get something to work partially.

xxyzz:
The proper way to enforce data structure is using pydantic, it would take more efforts than create types. But define typeddict is almost like define pydantic models.

And the first two commits are merge commits...

xxyzz:
You could try `python -m ruff format` to format code, I have removed black from dev dependency in the wikitextprocessor package but haven't remove it in this repo.

kristian-clausal:
wtf, those merges don't make sense!

This isn't about data structure, this is about type-checking the code.

xxyzz:
IMO add typeddict don't have much benefits compares to pydantic models. Good luck wrestle with mypy...

kristian-clausal:
I couldn't figure out a way to get rid of those merges except merge master to origin/master, push that and then rebase the branch to that (which fixed it).

This is why I've been avoiding rebase, it is such a mess. Merges may be messy, but they're clear and simple, while rebases, every time I've touched them, mess something up. Not a fan.

xxyzz:
That's not git rebase's fault. I suggest your branch is not created correctly.

kristian-clausal:
If I can't use a tool properly, *then I shouldn't be using that tool*! If it makes you any happier, I can just switch back to using merge.

xxyzz:
I'm trying to encourage you to try the rebase feature to avoid the situation like the last pr, I'm not even block that pr...

It's easier for me to say nothing.

kristian-clausal:
This is getting a bit too long, and most of the changes do not actually affect any behavior so I'll be merging this and will make a new branch in the future.

Issue #448, opened 2024-01-04
Output from subextractors need to be a bit more closely aligned with original output


kristian-clausal:
Tatu says: Because the output from the original extractor is being used with other projects that rely on it being "stable", changes to it need to be minimized, while all the outputs from all the projects need to be as close as possible.

I've started on trying to use our html-generation code to create websites from the data extracted with the other extractors, so I will be just posting here issues as they come along:

The first breaking difference is just that "lang_name" in word base data is different from "lang". Because "lang" is used in the original output and there is nothing *wrong* with it (not really, it is perfectly fine as is) the direction of change here is "lang_name" -> "lang". This change should be pretty simple.

I will continue with trying to make html generation work with the other extracted data and will post here stuff as things come along.

empiriker:
I totally agree with this sentiment.

However, I want to comment that initially when I started working on other extractors the consensus was that we shouldn't worry too much about enforcing a common output schema initially since we didn't know the requirements of the different Wiktionary editions. So I took some liberties introducing new fields and might not have been overly attentive to keeping everything aligned (though I tried).

Perhaps now with the json schemas and outputs of different extractors, it is a good time to revisit the issue and align the data structure as much as possible.

I want yet to advocate for allowing extractors to define new fields since some editions contain unique information or allow for a clean and more granular extraction than others.

For the extractors using pydantic, it could make sense to define base models that all extractors should/must inherit from. 

Good luck for your endeavors!

kristian-clausal:
I actually got our html generation code to generate a page from a sample of the Chinese data, and the only changes needed to make it work was:

- basically change all "lang_name" -> "lang". At first I only changed these in the base data for the word with `jq` (which took a while when I tried to figure it out), but then when the generation broke when translation items needed "lang" too I just used `sed` to blast the whole file, so there might have been a couple of other `lang_name` != `lang` conflicts somewhere else.

![Screenshot at 2024-01-04 11-14-28](https://github.com/tatuylonen/wiktextract/assets/93529179/803606e5-6b65-4aad-a64f-86dc7613788a)

Funnily enough, there was only one language in the Zh data sample I took (using `shuf` and then tailing that to get 10k entries) to have 500 entries so that it would end up on the front page of the dictionary... 世界語, Esperanto! Does Chinese Wiktionary have a really big Esperanto fanbase?

xxyzz:
[...]

PR #449, opened 2024-01-04
Use linkage model in Russian Wiktionary

empiriker:
This is in reply to #448 where @kristian-clausal mentions that the Russian extractor should use at least a basic Linkage model serializing to `{"word": linkage_word}`.

Mentioned in comment: https://github.com/tatuylonen/wiktextract/issues/448#issuecomment-1876986116_


xxyzz:
Thanks for your contribution!

kristian-clausal:
Thanks, I'll test it out later!

PR #452, opened 2024-01-05
Align schemas of Spanish, Russian and German extractors

empiriker:
In response to #448 .

While I am fine with all the suggested changes, I still want to point out that `lang_name` is objectively the better property name than `lang`, thanks to being more descriptive and unique. 

However, for the sake of uniformity, I am fine with renaming it.


kristian-clausal:
Thank, I will do more testing next week after the kaikki generation has run with the new stuff so that the data is more aligned.

PR #459, opened 2024-01-12
Russian edition: gloss data should be in list

kristian-clausal:
Just to be consistent with other extractors, raw_glosses and glosses should be lists of strings. This is in case there is a need to have hierarchical glosses (`["A person:", "Someone who persons."]`).


kristian-clausal:
If anyone's curious that I'm doing a specific change to these, it's because I want to use our html site generation code without doing too many changes to *that*, using it as a baseline. So when something breaks when generating the html, that's something that needs to be more closely aligned to how the original en.wiktionary json looks like.

PR #461, opened 2024-01-16
Remove optional types in Russian pydantic model

xxyzz:
Use Optional types would make the JSON schema harder to read.


empiriker:
I agree that there wasn't a consistent approach to optionality between the pydantic models and subsequently the json schemas for the different extractors. I also agree that it's more important to have a consistent approach here throughout the project than not.

However, I strongly advocate taking the rigorous approach here.

My argument is: The JSON schema isn't there to be primarily human readible, it's there to be precise and machine-readible. Ideally we would have both, but I think the latter is more important. (And since wiktextract is about machine-readible data, I think you would agree at least in spirit.)

Why is not specifying optionality not precise? The reason lies here:

https://github.com/tatuylonen/wiktextract/blob/master/src%2Fwiktextract%2Fextractor%2Fru%2Fpage.py#L254

This removes all keys from the output that have only been set with their default values. I assume we all agree that this is the behavior we want, i.e. not output a lot of keys to empty strings or null.

_Which by the way introduces another subtlety: whether the default value should be `None` or `""` for string values. The former has the advantage of allowing us to set values to empty strings without them being removed from the output. In same cases we might want that, in most this distinction will be irrelevant._

Why do I care about precise schemas? I'm coming from a, I think, typical use case: Let's say I take the output from wiktextract and want to use it in an application. Now the precise schemas are extremely helpful. I can use them directly, auto-convert them to Typescript types or use them to specify to a database schema fir wiktextract data. 

In all cases, I can then use precise type hints while developing my application on top of wiktextract data. And at this step it's crucial to know whether or not I can rely on a key to be present or not. Even if I should know to assume that most of our keys will not be present in a given output, it's much better when my IDE reminds me of that.

I'm happy to hear your thoughts and will be fine with whatever you decide in the end. I just want to advocate against removing specificity that might be useful to someone.

xxyzz:
I think only the fields without default value are required, all the rest fields with default value are optional, that's how the "required" list in JSON schema are generated. Shouldn't you use the "required" list?

empiriker:
Great! 

Then my mental image of how the schema gets created from pydantic models was incomplete and I was overly cautious. And my time would have been better spent just checking my mental image than drafting above comment... :innocent: 

No further comment then but thanks for improving readability!

PR #462, opened 2024-01-18
Remove Optional types in de and es pydantic models

xxyzz:
And use more strict configurations.


xxyzz:
The Python 3.12 test is slower is because the API coverage package uses is slower on 3.12, the test should be faster when coverage is not used.

PR #466, opened 2024-01-25
Extract forms table from expanded "прил*" template

xxyzz:
Inflection tables are so difficult to process...

### DIFF
--- a/src/wiktextract/extractor/ru/example.py
+++ b/src/wiktextract/extractor/ru/example.py
@@ -1,7 +1,8 @@
 from wikitextprocessor import WikiNode
-from wiktextract.extractor.ru.models import Example, Sense
 from wiktextract.page import clean_node
 from wiktextract.wxr_context import WiktextractContext
+
+from .models import Example, Sense
 
 EXAMPLE_TEMPLATE_KEY_MAPPING = {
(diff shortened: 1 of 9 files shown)


kristian-clausal:
They are a real pain, yeah. At some point, the table-parsing stuff in the main inflection parser needs to be refactored into its own bit, so that it can be used to parse other things other than tables (like those sound tables)...

xxyzz:
I also feel it's difficult to generalize the wikitext and html table process code, almost every table has its unique thing... like some tables use a data cell wikitext or td html tag for a table header.

kristian-clausal:
You were wondering about the "table header heuristics rejected" debug messages earlier this week, that's what those debug messages are for. Each language has a bunch of different explicitly declared header data that we use to determine whether a cell should actually be considered a header. It's a real mess... the debug messages are there to collect data on what is actually happening with the headers and are used to generate candidate data that is then later manually pruned to make whitelists.

xxyzz:
Couldn't we ask the Wiktionary editors to use the correct wikitext and html tag for table header? Maybe they would agree with the change.

kristian-clausal:
I am afraid to even try.

Issue #471, opened 2024-01-26
Chinese edition extraction contains entries with no part-of-speech


tatuylonen:
The Chinese edition extraction seems to contain 100000+ entries with no part-of-speech.

Also, it seems to contain translations with no language code.

xxyzz:
I think most of them are "redirect" pages that only have a template says for details see another word page, some examples: [100均](https://zh.wiktionary.org/wiki/100均), [別个](https://zh.wiktionary.org/wiki/別个), [chāi-tòng-chāi-tōe](https://zh.wiktionary.org/wiki/chāi-tòng-chāi-tōe)

I could add a "redirect" field for these pages.

xxyzz:
Expect for the "soft redirect" pages, I think most data that lacks pos or translation language code is from low quality pages that don't have POS title, use nonconvertible translation language names or have Lua/wikitext errors,

tatuylonen:
If the redirect pages are semantically same as the "hard" redirects in the English edition, then I would suggest encoding them in the same way.  But I assume there could be multiple redirect templates with these "soft" redirect pages.

Opinions include e.g.:
  - generate multiple "hard redirect" entries for the same word in the data
  - define a new "pos" field for these, e.g., "soft-redirect"
  - leave as is.
 
 I would prefer to have a clear "pos" field in each entry.  The existing "hard" redirects in English are currently an exception - but it might make sense to add a "hard-redirect" pos for them.  I would vote for adding "soft-redirect" pos field for the soft redirects, and "unknown" for the low-quality pages that have no part-of-speech indication.

Opinions?

xxyzz:
Sorry I missed your comment, I have created a draft pr #531 to change the pos fields.

xxyzz:
All check JSON errors are fixed for the zh edition on kaikki.org, close as resolved.

PR #476, opened 2024-01-29
Improve zh edition translation code

xxyzz:
There are too many translation list layouts in Chinese Wiktionary, this pr only covers part of them.

Issue #471


kristian-clausal:
I feel your pain... :pray:

xxyzz:
I guess it's because all non-English Wiktionary sites have way less active editors than the English Wiktionary, therefore there are many not well maintained pages.

PR #482, opened 2024-01-30
More other-edition field stuff

kristian-clausal:
Because I messed up my testing last time (by forgetting to use "--all" in wiktwords when generating the input json!!!!!), I need to do all the target language htmlgeneration attempts again, which will take a while but isn't complicated (now that I've figured out a procedure).


kristian-clausal:
Spanish wiktionary has some template calls that we assume have args, but don't; basically, just
```
=== Etimología ===
{{etimología}}
```

So in this case the change is more appropriate on the kaikki.org side of things...

PR #494, opened 2024-02-06
Fix bug introduced from 04ca7a8126e749f8fb500fdb96e43a8f8316d7ae

xxyzz:
This is why Python magic methods shouldn't be used, `not "False"` is `False`.


kristian-clausal:
Oops... :frowning:

PR #506, opened 2024-02-20
Create language names regex pattern used in `process_categories()` only once

xxyzz:
`get_all_names()` returns many language names, this change almost makes `test_long` test twice as fast on my machine.

But is this very long regex pattern really necessary?


kristian-clausal:
This is better than what I had in mind (got stuck in the weeds).

Issue #500, opened 2024-02-14
config.OTHER_SUBTITLES mixing strings and lists of strings


kristian-clausal:
I've been doing type annotations, and OTHER_SUBTITLES switching between strings in one source data file ('other_subtitles.json') and lists of alias strings another is messy, we should probably have better separation for that. The English extractor uses, for example, `wxr.config.OTHER_SUBTITLES["translations"]` as a string, assuming that it is a very specific term used in Translation sections and translation subpages, while the French one has a bunch of aliases and synonyms.

xxyzz:
The fr edition's title JSON files also have a different meaning, the data are the first parameter values of the [S](https://fr.wiktionary.org/wiki/Modèle:S) title template. Maybe we should change all values in the "other_subtitles.json" file to list?

kristian-clausal:
I think that lists should probably be kept for aliases, and 'definite' versions should be separate, in another file? `OTHER_SUBTITLES` is a bit too broad as a name, it doesn't say what it does. All the _SUBTITLES data collections are different, LINKAGE_SUBTITLES is a dict is "section name in language" -> "internal_name_for_section", POS_SUBTITLES is a dict where the values are dicts with fields containing data about the POS, and OTHER_SUBTITLES is a dict where the values are mixes of lists of strings and strings. Could use some cleaning up.

Other things: linkage_subtitles is "name in language" -> "our_internal_name_string", while other_subtitles is opposite, internal_name_string -> name. POS_subtitles is also "name" -> "internal_name".

xxyzz:
I don't think the code needs to know whether a string is alias, separate them to different files also requires combine them when loading the files.

I think most title strings are used in extractor folders and if they are not used outside they could be hard coded in extractor code.

xxyzz:
I think the title JSON files was added before we using separate extractors for different editions, and at that time the titles need to be loaded dynamically according to the dump file language code. But now because extractor code are loaded dynamically we don't need to do that. And all the title variables are indeed only used in extractor/* folders.

kristian-clausal:
That is very true, switching to consts should be pretty easy then.

kristian-clausal:
Closed as complete.

PR #512, opened 2024-02-22
Move de edition title JSON files to Python code

xxyzz:
Issue #500


xxyzz:
I'll update the en edition and remove the title variables in the `config` class tomorrow.

Only `OTHER_SUBTITLES` is used outside the extractor folder: https://github.com/tatuylonen/wiktextract/blob/efa31c523fcdcf8de9e394cc210fcf87203991e6/src/wiktextract/wiktionary.py#L56

this could be moved inside `parse_page()`, I think only en and zh editions have translation pages.

PR #525, opened 2024-02-28
Translate some common raw tags in fr edition code

xxyzz:
- Some raw tags weren't added because I'm not sure what's the equivalent tag, for example, "Passé composé", "Plus-que-parfait", "Passé antérieur", "Futur antérieur" in the conjugaison table: https://fr.wiktionary.org/wiki/Conjugaison:français/voir
- I didn't combine the tags because I'm afraid I might use the wrong order. At least the tags are tags now.


kristian-clausal:
"passé composé" is the past tense, except with a compound construction, and I guess we use "multiword-construction" for that, so "past multiword-construction". passé simple is just "past". En.wiktionary calls it "present perfect", but in this case just doing the one-to-one is good enough if you don't want to go with "present perfect". In this case, 'passé composé' is a very specific term in French for a French grammatical thing, and when you have that kind of thing in any language then "general" and "international" linguistic jargon goes out the window. If in doubt, just cheat by looking at en.wiktionary and seeing what it calls it.

There is no wrong order for tags, the tags are unordered, just combine them.

The "antérieur" correspond to "anterior".

xxyzz:
Thanks! I added the rest table headers and also translated some tags in translation list and form line template.

PR #528, opened 2024-03-04
Translate some fr edition gloss and form table tags

xxyzz:


xxyzz:
Most raw grammatical tags occur more than 10000 times should be translated, there are a few topic tags and sound locations yet to be translated.

PR #531, opened 2024-03-05
Make zh edition's pydantic model `WordEntry.pos` field required

xxyzz:
Default is `unknown`, and `soft-redirect` for `zh-see` and `ja-see` templates.


kristian-clausal (approved):
This all seems sensible to me! :+1:

PR #540, opened 2024-03-12
Urlstarts

kristian-clausal:
Some urls with `//` were passing through from modules etc. In Wikitext, the `//` is handled by replacing it with the current applicable URL prefix (typically `https://`) based on the page's url, but it's not really sensible for us so let's accept `//` as a prefix.

Some changes to either French inflection tests or to French inflection code was needed, if you could take a look @xxyzz

### DIFF
--- a/tests/test_clean.py
+++ b/tests/test_clean.py
@@ -348,3 +348,10 @@ class WiktExtractTests(unittest.TestCase):
         self.assertEqual(
             clean_node(self.wxr, None, tree.children), "2ちゃんねる, italic"
         )
+
+    def test_protocol_relative_url(self):
+        # https://en.wikipedia.org/wiki/Wikipedia:Protocol-relative_URL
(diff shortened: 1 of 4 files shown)


tests/test_clean.py
```diff
@@ -348,3 +348,10 @@ def test_bold_node_in_link(self):
         self.assertEqual(
             clean_node(self.wxr, None, tree.children), "2ちゃんねる, italic"
         )
+
+    def test_protocol_relative_url(self):
+        # https://en.wikipedia.org/wiki/Wikipedia:Protocol-relative_URL
```
kristian-clausal, line 353: I should have digged deeper about these. Because `//` is just deprecated (not forbidden), there will still be a lot of code that will use it -- and it is not necessarily true for every wikiproject that they have switched to hard `https://`.

PR #541, opened 2024-03-13
Remove category languages regex pattern

### LINKED ISSUE #537: Category "English phrasal verb" missing in "turn up"
The word "turn up" had, in an earlier version of the kaikki.org JSON outputs, the category "English phrasal verbs" in one of its senses.

Today, this category is missing from there, even though the Wiktionary entry still has the category:

https://kaikki.org/dictionary/English/meaning/t/tu/turn%20up.html
https://en.wiktionary.org/wiki/turn_up

Would be nice if this could be added again. Other phrasal verbs still have it:

https://kaikki.org/dictionary/English/meaning/r/ro/root%20around.html

xxyzz:
This huge regex pattern can't handle languages that have the same prefix("E" and "English"), split the category link by space and slash should serve the original code's purpose.

Fix #537

### DIFF
--- a/src/wiktextract/page.py
+++ b/src/wiktextract/page.py
@@ -7,19 +7,19 @@ from collections import defaultdict
 from copy import copy
 from typing import Any, Callable, Optional, Union
 
-from mediawiki_langcodes import get_all_names, name_to_code
+from mediawiki_langcodes import name_to_code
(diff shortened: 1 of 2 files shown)


kristian-clausal (approved):
Yeah, this is much more sensible.

PR #542, opened 2024-03-13
Use `Wtp.namespace_prefixes()` to check if a link is an image

xxyzz:
also replace the hard-coded strings in `repl_link()`


xxyzz:
The spaces before "File" are removed in these regex patterns: https://github.com/tatuylonen/wiktextract/blob/47f74c7e93d0cc20ffd93c729d2794172362ca24/src/wiktextract/clean.py#L1506-L1518

I guess we can't use `Wtp.parse()` at here?

PR #543, opened 2024-03-13
Link text cleaning regex change

kristian-clausal:
Previously, matches that were handled with `repl1` in the clean_values code for handling links would match [[File:foo.jpg]] when it shouldn't; added a colon to the negative match block to prevent this.


xxyzz:
Mind I push commit to this branch? I should check the group is not None.

PR #545, opened 2024-03-13
Handle tables inside tables in clean_value()

kristian-clausal:
I finally got this regex thing figured out. What
a nightmare.

Previously, if you had tables (`{| .. |}`) within
tables of wikitext, clean_value would not remove everything because it would do only one pass. We can't make this greedy, because then two separate tables would lead to the removal of everything between them too, so
negative lookahead is needed.

Do the usual "remove innermost table" thing until
no changes are detected, hope everything works out ok.

Q: Why do we remove tables, but not HTML `<table>...</table>` elements?
A: Good question! Because it's hard, and has been previously overlooked. A table will often just lead to garbled text when 'cleaned' from tags, etc. so it might be best to also remove all tables, and have the user handle them otherwise. XXX: remove HTML tables?

### DIFF
--- a/src/wiktextract/clean.py
+++ b/src/wiktextract/clean.py
@@ -1372,7 +1372,10 @@ def clean_value(
         return " ".join(args[i:])
 
     def repl_link(m: re.Match) -> str:
-        if m.group(1) is not None and IMAGE_LINK_RE.match(m.group(1)) is not None:
+        if (
+            m.group(1) is not None
+            and IMAGE_LINK_RE.match(m.group(1)) is not None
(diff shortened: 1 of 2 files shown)


xxyzz:
Why not remove the wikitext and html table nodes in `clean_node()` before passing the node text to `clean_value()`?

kristian-clausal:
Because clean_value is an independent function that should not rely on clean_node().

xxyzz:
But it's easier to remove the parsed node when we have this information and `clean_node()` is used more often.

PR #564, opened 2024-03-28
Simplify de editon's `page.py` and `gloss.py` file

xxyzz:
I also updated the [mediawiki_langcodes](https://github.com/xxyzz/mediawiki_langcodes/releases/tag/v0.2.3) package to include some German language names, but upload new package to pypi.org is [disabled today](https://status.python.org/incidents/dc9zsqzrs0bv).


xxyzz:
New mediawiki_langcodes package has been uploaded to pypi.org.

PR #572, opened 2024-04-08
Fix `JSONDecodeError` in file "overrides/zh.json"

xxyzz:
:facepalm:


kristian-clausal:
JSON would be about 10% better if it allowed those extra commas. :)

PR #574, opened 2024-04-09
Preliminary work on enabling not splitting links

kristian-clausal:
Previously we would just split on certain commas etc. even when a word or linkage contained them, creating extra partitions of words:

Hunde, die bellen, beißen nicht -> "Hunde", "die bellen", "beißen nicht".

This changes collects link data (visible link text strings where .alnum() == False) and passes it as a new parameter ultimately to split_on_comma_semi.

There, we use the links as part of the splitting regex, but ignore them when handling the iterations.

Also a couple of tests, but other tests fail.


xxyzz (approved):
TBH, I think the en edition's functions are too long and complicated, I usually don't fully understand the code...

kristian-clausal:
Yeah, it needs some refactoring.

PR #582, opened 2024-04-12
Fix pydantic errors and remove empty glosses for zh edition

xxyzz:


kristian-clausal:
Thanks, sorry for the commit spam, you were much faster with the Pydantic update than I anticipated.

xxyzz:
I can't believe pydantic can't find this `tuple` and `list` type validation error until the latest version...

PR #587, opened 2024-04-16
Ignore "Statistics" sections to suppress "fi-stats-surname" Lua error

xxyzz:
en edition templates "fi-stats-given-name" and "fi-stats-surname" put the end of wikitext link `]]` inside `#switch` parser function break our parser.

Error: https://kaikki.org/dictionary/All%20languages%20combined/errors/details-Invalid-source-given-for-fi-stats-surn-NnZNGQei.html
Template pages:
- https://en.wiktionary.org/wiki/Template:fi-stats-surname
- https://en.wiktionary.org/wiki/Template:fi-stats-given-name


kristian-clausal (commented):
This seems to be the best way.

PR #590, opened 2024-04-16
Switch from root Logger in logging to instance Logger

kristian-clausal:
Because we were using logging.debug, logging.warning and logging.error, our logs are spammed by dependencies that shouldn't be sending debug messages (datetime, apparently).

Hopefully by switching to a instance of Logger instead using root, we can stop those messages.


kristian-clausal:
I've changed the levels of the datetime and dateparser modules in wikitextprocessor.

PR #594, opened 2024-04-18
Fix issue with untagified tags

kristian-clausal:
General PR for issues with tags.


kristian-clausal:
Instead of doing things piecemeal all over the code, we can try validating tags inside data_append.

PR #601, opened 2024-04-22
Pass --quiet to Wtp.__init__()

xxyzz:
new `quiet` argument is added in https://github.com/tatuylonen/wikitextprocessor/pull/273


xxyzz:
Some outdated language code template argument(`"psu" not valid`) caused Lua errors in zh edition are fixed on Wiktionary but weren't included in the 20240420 dump file, should be fixed in the next month's dump file.

PR #602, opened 2024-04-23
Get es edition's translation language name from expanded template

xxyzz:
I'll update the mediawiki_langcodes package to add language names and codes in the es edition in case we need these data later.


xxyzz:
es edition language data are added to mediawiki_langcodes [v0.2.6](https://github.com/xxyzz/mediawiki_langcodes/releases/tag/v0.2.6)

There are many Lua errors in zh edition's logs, most errors are already fixed but the changes weren't included in the current dump file, next week's dump file should have less Lua errors.

Issue #604, opened 2024-04-24
English translation of examples sometimes missing, included in the original text


29jm:
Some examples in some Czech words have an issue, where instead of having both `text` and `english` in the example object, there's only `text` containing a concatenation of both, e.g. "the original text ― the translated text".

Some examples:

- [ani](https://en.wiktionary.org/wiki/ani#Czech): only the last example has the issue
![_screenshot](https://github.com/tatuylonen/wiktextract/assets/6154077/02c1629b-929e-49ca-be14-2fef144e33f3)
- [dobrý](https://en.wiktionary.org/wiki/dobr%C3%BD#Czech): same
![_screenshot](https://github.com/tatuylonen/wiktextract/assets/6154077/35913e4b-112b-451d-bfc0-bc28da8e63b9)
- It's not necessarily for the last entry though, for instance with zničit.
- Some more like this: vzít, hlavní, čas, žít, paní, jméno, smrt, kniha, psát, názor, světový, osobní, minulý, onen, umění, věk, telefon, zástupce, ženský.

If it's a problem in the page markup, I can fix it there, but I didn't see what could cause it.

kristian-clausal:
Yeah, this is buggy, I'll take a look at it.

kristian-clausal:
Sorry, I got stuck trying to figure out a bug? with our logging system, this might have to wait a while.

29jm:
No worries at all, I know how it is :)

kristian-clausal:
Found part of the issue, and it's a silly one; the examples I looked at (ani and vzít) had a word that was blacklisted from our  "what is an English word" set: `He`, with a capital. Our function that determines what kind of language or function a string has rejects the strings because 1/4 of it is classified as non-english, so it all gets lumped as a `romanization` string and the example is concatenated.

Actually looking at vzít, the example that is broken (please provide more examples in the future so that I don't have to hunt for them) doesn't get through because it's 2/3rds Czech names with one English word.

I am writing this message as I'm going through example, and hlavní is a completely *different* issue: the example isn't in a template that we accept as an 'example' (`ux`) template, but as a `coi` template, related to `collocation`. Outputs the same as an example template, and outputs the same kind of stuff (looks like an example, quacks like an example), so I just added the template and its aliases to our list of "ux" templates.

I'm going to commit these, and hopefully most of all of the issues will be addressed. If you find that some weren't fixed, just point them out.

[...]

PR #615, opened 2024-05-06
Use example template args to determine example

kristian-clausal:
Previously, we would expand everything with clean_node and use heuristics to check whether it was an example or not.

However, seeing as how we have access to template
arguments, we can bypass some of these heuristics
early by checking if the text we're checking has
one template, from the example templates set,
with arguments that are the same as what is part
of the clean_node output. In this case, we can just assume this is an example.

### DIFF
--- a/src/wiktextract/english_words.py
+++ b/src/wiktextract/english_words.py
@@ -1874,6 +1874,12 @@ additional_words = set(
         "zealotry",
         "zoospores",
         "zygosperm",
+        "chamomile",
+        "peppermint",
+        "x-axis",
+        "y-axis",
+        "z-axis",
+        "maté",
     ]
 )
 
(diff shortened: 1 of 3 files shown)


kristian-clausal:
Meant to address #604

xxyzz:
I notice en edition code doesn't handle the [zh-x](https://en.wiktionary.org/wiki/Template:zh-x) template properly, this template displays the example text in Traditional Chinese and Simplified Chinese forms, and could also be used in etymology section. Example page: [未雨綢繆](https://en.wiktionary.org/wiki/未雨綢繆)

zh edition code parses the HTML tags expanded from the template. zh edition's template generates the same HTML tags as the en edition, but because zh code uses pydantic, the code need some changes for en code, please see #613

kristian-clausal:
I introduced some exception-triggering bugs with this, so I want to first fix that and then leave it until next week (because I'm taking Friday off and Thursday is a holiday); I'll take a look at the Chinese template next week.

PR #628, opened 2024-05-13
Zhexamples

kristian-clausal:
I've added one more branch to the extract_example multi-line if-tree to account for how `Template:zh-x` formats its examples. Because it's a special case, it's simpler than usual (just using `classify_desc` to put each line into a box).

For example in: https://en.wiktionary.org/wiki/%E6%9C%AA%E9%9B%A8%E7%B6%A2%E7%B9%86

Note that each line is split on square brackets because `[` interferes with `classify_desc`, most probably on purpose (it classifies it as 'other' instead of 'romanization'). `zh-x` uses a lot of "text [Specific Chinese language, trad. or simp.]" style qualifiers, which we're going to ignore and add as part of the text itself. Examples inside senses don't have a `tags` field, and I don't want to add them unless there's a lot more need for it.

The `[` also broke the heuristics for the original code, so I also added a negative condition regarding `zh-x` up in the first if condition to counteract that. If there are more Chinese templates like this, or maybe if all Chinese examples have this exact format, then we can expand the condition or make it more general (with "lang_code == 'zh'", for example).

It took me too long messing with this code to realize it would be a royal mess to integrate this stuff into the 'general' branches, and adding a special case isn't going to be that expensive. Just makes the code even longer.

### DIFF
--- a/src/wiktextract/extractor/en/page.py
+++ b/src/wiktextract/extractor/en/page.py
@@ -550,6 +550,13 @@ ignored_etymology_templates: list[str] = [
     "cln",
     "langname-lite",
     "no deprecated lang param usage",
+    "mention",
+    "m",
+    "m-self",
+    "link",
+    "l",
+    "ll",
+    "l-self",
 ]
 # Regexp for matching ignored etymology template names.  This adds certain
 # prefixes to the names listed above.
--- a/src/wiktextract/extractor/en/page.py
+++ b/src/wiktextract/extractor/en/page.py
@@ -821,13 +821,21 @@ def parse_language(
         data_append(base_data, "tags", "reconstruction")
     sense_data: SenseData = {}
     pos_data: WordData = {}  # For a current part-of-speech
-    zh_pron_data: WordData = {}  # Chinese Pronunciation-sections in-between
+    level_four_data: WordData = {}  # Chinese Pronunciation-sections in-between
     etym_data: WordData = {}  # For one etymology
     pos_datas: list[SenseData] = []
-    zh_pron_datas: list[WordData] = []
+    level_four_datas: list[WordData] = []
     etym_datas: list[WordData] = []
     page_datas: list[WordData] = []
     have_etym = False
+    inside_level_four = False  # This is for checking if the etymology section
+    # or article has a Pronunciation section, for Chinese mostly; because
+    # Chinese articles can have three level three sections (two etymology
+    # sections and pronunciation sections) one after another, we need a kludge
+    # to better keep track of whether we're in a normal "etym" or inside a
+    # "level four" (which is what we've turned the level three Pron sections
+    # into in the fix_subtitle_hierarchy(); all other sections are demoted by
+    # a step.
     stack: list[str] = []  # names of items on the "stack"
 
     def merge_base(data: WordData, base: WordData) -> None:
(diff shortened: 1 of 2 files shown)


src/wiktextract/extractor/en/page.py
```diff
@@ -3538,7 +3539,9 @@ def usex_template_fn(name, ht):
                             lines = [parts[0].strip()]
                             tr = parts[1].strip()
                 elif len(lines) > 1:
-                    if any(re.search(r"[]\d:)]\s*$", x) for x in lines[:-1]):
+                    if any(
+                        re.search(r"[]\d:)]\s*$", x) for x in lines[:-1]
+                    ) and example_template_names in (["zh-x"], ["zh-usex"]):
```
xxyzz, line 3544: Should be `not in`.

This might be difficult, but I think it's better to put the "zh-x" code in a new function so it could be reused for  etymology section. And process the expanded HTML tags would fix the unreliable `classify_desc` problem.

xxyzz:
The "etymology_text" field is still missing for the Chinese section JSON data in page "作", there might be a bug in code adding the extracted etymology data to the final dictionary variable.

src/wiktextract/extractor/en/page.py
kristian-clausal: Thanks for spotting that, I was hasty.

The classify_desc problem isn't with HTML tags, it's because classify_desc breaks in square brackets, like "[Chinese language name, trad.]".

kristian-clausal:
I tested this with a stripped-down version of the article, but didn't try the the whole thing. Using the whole Chinese section, there's not etymology data.

src/wiktextract/extractor/en/page.py
xxyzz: I mean if we could processing HTML tags then separating the text would be easier. Because language variety and other texts can be distinguished through HTML tag name and CSS class names. This is just a suggestion, if we want to move the data inside brackets to tags in the future, we could copy the zh edition code, they use the same HTML tags.

kristian-clausal:
The issue with 作 seems to be that Glyph origin, Etymology and Pronunciation are all on the same level. My minimal zuo.txt didn't have the Pronunciation titles + pron subsections, which is why the etymologies got associated with its definitions.

Chinese articles look like this:

```
===Glyph Origin===
...
===Etymology===
...
===Pronunciation 1===
...
====Definitions====
...
===Pronunciation 2===
...
====Definitions====
```

Currently it seems etymology information for POS sections is not duplicated and the first POS section on the page receives it. 

Ok, I found a Chinese article where the Pronunciation is a level 4 `====Pronunciation====`.

A hack might be to make the Pronunciation section a level 4 when it's right after an Etymology template. The other level 4 sections after that (under the Pronunciation template previously) will now be children of the Etymology node, and other Pronunciation sections will be left alone.

xxyzz:
The etymology section data are added now but the first pronunciation section data are added to the second "definitions" POS section data.

xxyzz:
All issues seem to be fixed. Should we merge this, or you want to test the code on more pages?

kristian-clausal:
I don't trust this *at all*, I had so much trouble getting it to work, so I will try to diff (or whatever is appropriate for json) differences in output between the main branch and this branch. Thanks for taking a look!

kristian-clausal:
I FINALLY got the jsondiff thing to work. The script is slow as molasses, I got caught trying to figure out why things didn't work due to several minor bugs compounding each other (for example: a `continue` that was indented too low, and having the name of the file be 'jsondiff' which messed up with the importing of jsondiff...), then I had messed up because I needed to re-extract stuff due to unrelated changes to the code (ignored etymology templates causing diffs, of course)... But looking at the diff right now, and adding some new section terms to a couple of places, it seems this is where I want it to be!!!

PR #640, opened 2024-05-21
[de] extract adj form tables in "Flexion" namespace pages

xxyzz:


xxyzz:
de edition 20240520 dump file lacks the page text of "Modul:ErsetzeZeichen" Lua module, this will cause some Lua errors.

PR #645, opened 2024-05-24
Don't remove namespace prefix in a link from `clean_value()`

xxyzz:
Should only remove "w:" and "s:" prefixes.

Found this wikitext in https://de.wiktionary.org/wiki/Flexion:sehen , remove the namespace causes the de edition code add an extra form data: #644


kristian-clausal:
Sorry, I didn't notice you asked for a review! This looks fine to me.

kristian-clausal (approved):

PR #674, opened 2024-06-14
Parse Template:+obj and add new `coincidence` field

kristian-clausal:
The `Template:+obj` is used to generate formatting for constructions such as `[+accusative]` or
`[+foo = a meaning with bar]`, used to show that something is usually coincident or tied to an argument or
a specific word or something else.

These were not previously handled in parsing, which poisoned `forms` with 'canonical' entries like
`pro [+accusative]`.

Instead of creating new categories of tags like
`with-accusative` etc., Tatu gave the green light
to make a completely new field, which is here called "coincidence" because it's about coincident... things.

`coincidence` is a dict with a `tags` field, `words` field and `meaning` field (each optional).

Currently only used at the main entry level for
when a word head uses `+obj`; see pro/Czech.

### DIFF
--- a/src/wiktextract/form_descriptions.py
+++ b/src/wiktextract/form_descriptions.py
@@ -18,9 +18,9 @@ from typing import (
 import Levenshtein
 from nltk import TweetTokenizer  # type:ignore[import-untyped]
 
-from wikitextprocessor.core import TemplateArgs
 from wiktextract.type_utils import (
     AltOf,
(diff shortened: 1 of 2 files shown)


kristian-clausal:
After discussing with Tatu, instead of doing a specific kludge for `Template:+obj`, we're going to create a new "extras" or "info_templates" field that is similar to `etymology_templates`, and use it in root data and sense data.

Issue #676, opened 2024-06-15
zh-form template missing in extract for Chinese words


jiong3:
The form template seems to be missing from the Chinese entries, e.g. https://en.wiktionary.org/wiki/%E6%A9%9F%E8%BB%8A would have the following zh-form template:
{{zh-forms|s=机车}}
; “motorcycle”
: {{zh-etym-short|機器腳踏車}}.

I would like to use this information to know the simplified version (s=) of the word, if there is one.

xxyzz:
This is fixed on kaikki.org, close as resolved.

jiong3:
Great, thank you!

PR #677, opened 2024-06-17
[en] extract forms data and literal meaning from "zh-forms" template

xxyzz:
The code is modified from the zh edition code. The new function adds two tags("Simplified Chinese", "Traditional Chinese") and a new "literal_meaning" JSON field. The literal meaning data is commonly provided in Chengyu pages.

Data are added to `base_data` because "zh-forms" is above POS sections and should be included to all the following POS sections data.

Since the code are mostly the same as the zh edition code, tests are not added.

Example pages:
- https://en.wiktionary.org/wiki/未雨綢繆  (Chengyu, has `lit` parameter)
- https://en.wiktionary.org/wiki/僥倖  (has two POS sections)

GitHub issue #676

### DIFF
--- a/src/wiktextract/extractor/en/page.py
+++ b/src/wiktextract/extractor/en/page.py
@@ -52,11 +52,12 @@ from wiktextract.translations import parse_translation_item_text
 from wiktextract.type_utils import (
     DescendantData,
     ExampleData,
+    FormData,
+    LinkageData,
     SenseData,
     SoundData,
-    WordData,
-    LinkageData,
     TemplateData,
+    WordData,
 )
 from wiktextract.wxr_context import WiktextractContext
 
@@ -3331,10 +3332,11 @@ def parse_language(
             if not isinstance(node, WikiNode):
                 # print("  X{}".format(repr(node)[:40]))
                 continue
-            if isinstance(
-                node, TemplateNode
-            ) and process_soft_redirect_template(wxr, node, redirect_list):
-                continue
+            if isinstance(node, TemplateNode):
+                if process_soft_redirect_template(wxr, node, redirect_list):
+                    continue
+                elif node.template_name == "zh-forms":
+                    process_zh_forms_templates(wxr, node, base_data)
 
             if node.kind not in LEVEL_KINDS:
                 # XXX handle e.g. wikipedia links at the top of a language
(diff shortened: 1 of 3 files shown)


xxyzz:
"zh-forms" usually used directly under the language title. https://en.wiktionary.org/wiki/機車 seems to be an outlier.

src/wiktextract/extractor/en/page.py
```diff
@@ -4190,3 +4192,43 @@ def process_soft_redirect_template(
...
+            }
+            if len(form_data["form"]) > 0:
+                base_data["forms"].append(form_data)
+        elif p_name == "alt":
+            for form_text in clean_node(wxr, None, p_value).split(","):
+                texts = form_text.split("-")
+                form_data: FormData = {"form": texts[0]}
+                if len(texts) > 1:  # pronunciation data
+                    form_data["raw_tags"] = texts[1:]
+                if len(form_data["form"]) > 0:
```
kristian-clausal, line 4227: Is this for a single hanzi character?

src/wiktextract/extractor/en/page.py
```diff
@@ -4190,3 +4192,43 @@ def process_soft_redirect_template(
...
     return False
+
+
+def process_zh_forms_templates(
+    wxr: WiktextractContext,
+    template_node: TemplateNode,
+    base_data: WordData,
+) -> None:
+    # https://en.wiktionary.org/wiki/Template:zh-forms
+    base_data["forms"] = []
```
kristian-clausal, line 4203: Is it guaranteed that base_data["forms"] doesn't already have any data? Maybe this should be a simple check:

```
if not base_data["forms"]:
    base_data["forms"] = []
```

Just in case the situation changes in the future and someone adds code that populates base_data before process_zh_forms_templates is called.

kristian-clausal (changes requested):
My only quibble is the unconditional `base_data["forms"] = []`, which could bite us in the future if something is added that happens before process_zh_forms_templates.

src/wiktextract/extractor/en/page.py
xxyzz: Some page could put dialect information in the "alt" parameter, for example `alt=新抱-Cantonese` in page https://en.wiktionary.org/wiki/新婦

kristian-clausal (approved):

src/wiktextract/extractor/en/page.py
kristian-clausal: I misread the line, I thought it was "if len(texts) > 0". :+1:

PR #708, opened 2024-07-05
More cleaning

kristian-clausal:
Fixing annotations, mostly.

I've looked a bit at mypy output for the other extractors, and it's helped to find issues with annotations in wikitextprocessor and elsewhere in wiktextract (for example adding some `@overload` stuff). @xxyzz do you mind if I start poking around in there? It should be non-destructive (hopefully).

### DIFF
--- a/src/wiktextract/english_words.py
+++ b/src/wiktextract/english_words.py
@@ -6,8 +6,8 @@
 #
 # Copyright (c) 2020-2022 Tatu Ylonen.  See file LICENSE and https://ylonen.org
 
-import nltk
-from nltk.corpus import brown
+import nltk  # type: ignore[import-untyped]
+from nltk.corpus import brown  # type: ignore[import-untyped]
 
 from .form_descriptions_known_firsts import known_firsts  # w/ our additions
 
--- a/src/wiktextract/extractor/de/section_titles.py
+++ b/src/wiktextract/extractor/de/section_titles.py
@@ -1,7 +1,7 @@
 from ...config import POSSubtitleData
 
 # argument of title template https://de.wiktionary.org/wiki/Vorlage:Wortart
-POS_SECTIONS: POSSubtitleData = {
+POS_SECTIONS: dict[str, POSSubtitleData] = {
     "Abkürzung (Deutsch)": {"pos": "abbrev", "tags": ["abbreviation"]},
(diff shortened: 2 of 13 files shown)


kristian-clausal:
Leaving this uncommitted until next week, not a high priority.

xxyzz:
I haven't use mypy for a while, I think pydantic and ruff gives more useful suggestions. Maybe mypy could find more meaningful errors in code not using pydantic(I doubt it)?

And frankly, I think it's overkill for type annotations longer than the actual code...

src/wiktextract/english_words.py
```diff
@@ -6,8 +6,8 @@
 #
 # Copyright (c) 2020-2022 Tatu Ylonen.  See file LICENSE and https://ylonen.org
 
-import nltk
-from nltk.corpus import brown
+import nltk  # type: ignore[import-untyped]
```
xxyzz, line 9: Should ignore "nltk" imports in mypy config file.

xxyzz:
And the "whitespace at end of line" ruff warning is a legit suggestion: https://docs.astral.sh/ruff/rules/blank-line-with-whitespace

kristian-clausal:
> And the "whitespace at end of line" ruff warning is a legit suggestion: https://docs.astral.sh/ruff/rules/blank-line-with-whitespace
> 
> The [.editorconfig file](https://github.com/tatuylonen/wiktextract/blob/master/.editorconfig) was added to let the editor removes trailing whitespace.

This isn't about blank lines with whitespace, this is whitespace *at the end* of a line, which is triggered *inside quotes*, in a string. This is obviously wrong.

```
"""
This space triggers the warning -> 
"""
```

xxyzz:
Please read the doc again...

kristian-clausal:
You linked to W293, the error I am talking about is **W291**.

kristian-clausal:
The ignore is also just a `# noqa:` ignore at the end of the string itself. Please read more carefully.

xxyzz:
Both rules are created for the same reason(no trailing whitespace anywhere defined in PEP8)... If you think it's bug, you could create a bug report to the ruff project.

kristian-clausal:
It's already been fixed in upstream ruff.

PR #714, opened 2024-07-09
[fr] find etymology POS title text inside italic node

xxyzz:
Usually inside a link, but texts are also used.


xxyzz:
All exceptions in zh edition are fixed.

PR #715, opened 2024-07-09
In examples, "taxonomic" is the same as "english"

kristian-clausal:
Issue #604, Czech translations (continued)

In translations like

```
ví ucho ― Leonotis nepetifolia (literally, “lion's ear”)
```

the translation part starting with "Leonotis" is has its classification returned as "taxonomic" due to the heuristics used in classify_desc().

I've been trying to kludge something better here, but for this specifically the right call to make is to change it so that if a description is either "english" or "taxonomic", that counts as English. There is not meaningful distinction here in the examples when trying to figure out translation stuff.

The heuristics could be better, which is what I tried to figure out, but it works fine for now...

### DIFF
--- a/src/wiktextract/english_words.py
+++ b/src/wiktextract/english_words.py
@@ -1918,6 +1918,17 @@ not_english_words_1 = set(
         "novo",
         "pronto",
         "que",
+        "extortionary",
+        "democracy",
+        "Mrs",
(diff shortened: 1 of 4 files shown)


kristian-clausal:
I hate list and string manipulation math.

PR #716, opened 2024-07-10
[zh] extract data from low quality German pages

xxyzz:
Most "no-gloss" tag pages really don't have gloss at all...


kristian-clausal:
The bigger the edition, the more annoying it is because there are so many ways things can go wrong.

The smaller the edition, the more annoying it is because there is not enough good data...

And when it's  middle-sized, that's the worst of both worlds!

xxyzz:
In short, wikitext is annoying...

PR #758, opened 2024-08-07
Fix exceptions caused by None passed to unicodedata.normalize

kristian-clausal:
92 exceptions is not a lot, but enough to stop
the kaikki regen.

This call to unicodedata.normalize() seems to be
the only one that could have had None (others
are either in filtering `if` blocks or would
cause an exception due to the out of bound string
indexing).

I tried to find a final 93rd exception in extract.log (because I got that from `wc -l`) and it was the line "TOO MANY EXCEPTIONS DETECTED"... Which is weird, shouldn't this be in regenerate.log?

### DIFF
--- a/src/wiktextract/inflection.py
+++ b/src/wiktextract/inflection.py
@@ -625,7 +625,7 @@ def extract_cell_content(
             if ref:
                 deflst.append((ref, col[ofs : m.start()].strip()))
             ref = unicodedata.normalize(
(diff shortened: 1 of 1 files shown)


kristian-clausal:
Nevermind the extract.log, the kaikki `regen` script just outputs the same echo into both regenerate.log and extract.log at the end. The counting of exceptions happens before this, so the count (which reads extract.log) does not get polluted by the error message, all is fine.

PR #759, opened 2024-08-07
Remove `<nowiki/>` in `clean_value()`

xxyzz:
This tag could break HTML tag regex patterns below this line.


kristian-clausal (approved):
If there is a more correct way to do this, I don't know what it is.

Issue #763, opened 2024-08-08
raw_glosses and glosses fields


rdurkacz:
Here is part of the French extract for the word "livrer"

        "raw_glosses": [
          "(reflexive) abandon oneself, give oneself over [with à ‘to’]"
        ],
        "glosses": [
          "abandon oneself, give oneself over [with à ‘to’]",
          "abandon oneself, give oneself over"
        ],
        "tags": [
          "reflexive"
        ],

There are two overlapping glosses produced from one raw gloss. Presumably the reason for that is that it is not known how to deal with the text in [...] so that the user is offered the choice, keep or discard the [...]. Is there some logic by which the user can deal with this case automatically, perhaps always taking the first alternative along with the tags fields, so as not to lose or dupicate information?

kristian-clausal:
The two lines in glosses are meant to be used for when you have a situation like this:

```
# To a reversed order; half round; facing in the opposite direction; from a contrary point of view. {{defdate|from ca. 1350–1470<ref name=SOED/>}}
...
## {{lb|en|nautical}} To the opposite [[tack]]: see {{l|en|go about}}. {{defdate|from late 15th c.<ref name=SOED/>}}
```

Where you've got a hierarchical gloss.

```
      "glosses": [
        "To a reversed order; half round; facing in the opposite direction; from a contrary point of view.",
        "To the opposite tack: see go about."
      ],
```

The code is a big spaghettuous, so I can't make any guarantees other than intent.

In this case, I would say it's a bug. The raw_gloss text is added directly because we wanted a higher-level gloss, but when the subglosses are cleaned up (which the high-level gloss apparently isn't) then when the gloss is added it doesn't notice it's "already there" when we do a `gloss in data["glosses"]` check. This doesn't seem to have anything to do with the recent addition of handling for certain templates (which means, handling of Template:+obj), it's just that `+obj` outputs something the we remove in the gloss (the final `[brackets]`.

I'll see what I can do.

kristian-clausal:
This should be fixed, if you find any oddities later just make a new issue (or if it's specifically this just comment here and I'll reopen).

PR #764, opened 2024-08-09
Fix raw_gloss added directly to `glosses`

kristian-clausal:
This should really be looked at again.

Previously, we would add 'duplicate' lines into
glosses when the raw_gloss added and the extra-cleaned gloss handled when `subglosses` (even when there were no other glosses other than the first) were handled did not match. So we got a gloss that is a bit more raw followed by the same gloss, except slightly more
cleaned. This fix is a dumb one that *tries* not
cause other issues by simply checking if we're
adding the first subgloss and replacing the
raw_gloss with it completly... Which begs the
question, why is the raw_gloss added at all?

Aaah, commit text, what a rubber duck you are. I just got it; the raw_gloss is added before cleaning
as a super-gloss because it is needed in sense_base before subglosses are handled, so because the
subsequent glosses are cleaned in there,
the raw_gloss was only handled a little bit.


kristian-clausal:
This was such a pain. I spent much too much time trying to make this a function and to refactor the code somehow, but no, it's too tied up by side-effects that that wasn't possible, so I went with brutishly just copying the code dealing with string clean-up so that supergloss strings at least look presentable. The logic that fixed #763 is mainly the same (check index, add gloss) as in the earlier commits in this PR, just better executed.

PR #780, opened 2024-08-21
[pl] extract "odmiana-rzeczownik-polski" template parameters

xxyzz:
This template could use tag template in arguments, previous code expands template to HTML table but it's tricky to distinguish word and tag.


xxyzz:
Page [Urban](https://pl.wiktionary.org/wiki/Urban) has parts of "ref" HTML tag extracted as forms, should be fixed when https://github.com/tatuylonen/wikitextprocessor/issues/299 is resolved.

PR #785, opened 2024-08-23
[ja] improve linkage, sound, translation, pos sections code

xxyzz:
Please notice I changed the wikitext link regex in the last commit to match file links in ja edition.


kristian-clausal:
Using `\w\d` is obviously better, in retrospect, for that. We don't live in the ASCII ages anymore.

xxyzz:
Failed? How?

kristian-clausal:
Wikitextprocessor?

xxyzz:
I think the compiled regex pattern is cached with another edition's namespace data from other tests.

PR #796, opened 2024-08-29
Add "adnominal" as POS

joshuawu:
Some Japanese entries depend on this, such as [この](https://en.wiktionary.org/wiki/%E3%81%93%E3%81%AE).

### DIFF
--- a/src/wiktextract/extractor/en/section_titles.py
+++ b/src/wiktextract/extractor/en/section_titles.py
@@ -21,6 +21,7 @@ POS_TITLES: dict[str, POSSubtitleData] = {
     "adjective": {"pos": "adj"},
     "adjectuve": {"pos": "adj", "debug": "misspelled subtitle"},
     "adjectives": {"pos": "adj", "debug": "usually used in singular"},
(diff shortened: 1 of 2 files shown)


kristian-clausal:
Thank you for your contribution! Everything seems correct.

Just in case someone wants to do a similar kind of commit in the future, the steps taken here were:

* Add an entry to `src/wiktextract/extractor/en/section_titles.py` and `src/wiktextract/part_of_speech.py`
* If the part-of-speech title you're adding fits into a pre-existing category, like `noun`, use that as its `pos` value
* But if it makes sense to add a new one, just add a new `pos` value into the above; we'll take a look at it, and in this case it makes sense to add `adnominal`, but putting it under `adjunct` would also make sense.
* Then if it seems like it's a specific kind of PoS that needs to be distinguished, like if we did add "adnominal" with `"pos": "adjunct"` then add a tag or two; each tag a separate string inside the `tags` field list. However, tags for the English edition of wiktionary must exist in `tags.py`, so check if you can find it there in `valid_tags`. Tags can be added easily, but it's unlikely you need to.

PR #808, opened 2024-09-09
[en] extract "zh-x" example templates properly

xxyzz:
Most code are copied from zh edition code with some modifications.

Separate Traditional Chinese, Simplified Chinese, dialect and other tag properly.

Also handle the common "zh-x" and "ja-x" template under "quote-*" template layout.

Example pages: [大家](https://en.wiktionary.org/wiki/%E5%A4%A7%E5%AE%B6), [尋釁滋事](https://en.wiktionary.org/wiki/%E5%B0%8B%E9%87%81%E6%BB%8B%E4%BA%8B)

The code could also be reused for example templates in Chengyu pages, like [買櫝還珠](https://en.wiktionary.org/wiki/%E8%B2%B7%E6%AB%9D%E9%82%84%E7%8F%A0)

### DIFF
--- a/src/wiktextract/extractor/en/example.py
+++ b/src/wiktextract/extractor/en/example.py
@@ -5,7 +5,7 @@ from wikitextprocessor.parser import NodeKind, TemplateNode, WikiNode
 
 from ...page import clean_node
 from ...tags import valid_tags
-from ...type_utils import ExampleData
+from ...type_utils import ExampleData, SenseData
 from ...wxr_context import WiktextractContext
 from ..ruby import extract_ruby
 
@@ -13,6 +13,7 @@ from ..ruby import extract_ruby
 def extract_example_list_item(
     wxr: WiktextractContext,
     list_item: WikiNode,
+    sense_data: SenseData,
     parent_data: Optional[ExampleData],
 ) -> list[ExampleData]:
     examples = []
@@ -22,6 +23,7 @@ def extract_example_list_item(
                 extract_template_zh_x(
                     wxr,
                     template_node,
+                    sense_data,
                     parent_data
                     if parent_data is not None
                     else ExampleData(raw_tags=[], tags=[]),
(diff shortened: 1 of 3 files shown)


src/wiktextract/extractor/en/example.py
```diff
@@ -0,0 +1,237 @@
...
+                    roman = clean_node(wxr, None, span_tag)
+                    for span_tag in dd_tag.find_html_recursively("span"):
+                        span_text = clean_node(wxr, None, span_tag)
+                        if span_text.startswith("[") and span_text.endswith(
+                            "]"
+                        ):
+                            roman_raw_tags.append(span_text.strip("[]"))
+
+        example_text = ""
+        last_span_is_exmaple = False
```
kristian-clausal, line 150: Minor typo in `last_span_is_example`.

src/wiktextract/extractor/en/example.py
```diff
@@ -0,0 +1,237 @@
...
+                    results.append(example)
+
+    # no source, single line example
+    if not has_dl_tag:
+        roman = ""
+        raw_tags = []
+        for span_tag in expanded_node.find_html(
+            "span", attr_name="lang", attr_value="Latn"
+        ):
+            roman = clean_node(wxr, None, span_tag)
```
kristian-clausal, line 182: Maybe add `break` or a comment explaining why this for loop "overwrites" `roman`; I know there's only one `span` coming from the generator, but it looks a bit strange at first sight.

src/wiktextract/extractor/en/example.py
```diff
@@ -0,0 +1,237 @@
...
+}
+
+
+def clean_example_empty_data(data: ExampleData) -> ExampleData:
+    raw_tags = data.get("raw_tags", [])
+    new_raw_tags = []
+    for raw_tag in raw_tags:
+        if raw_tag in valid_tags:
+            data["tags"].append(raw_tag)
+        elif raw_tag in ZH_X_TAGS:
```
kristian-clausal, line 226: This might be better first; if someone adds "trad." or "simp." to tags.py, this branch would break, and no one would know.

kristian-clausal (approved):

src/wiktextract/extractor/en/example.py
kristian-clausal: I mean, `elif raw_tag in ZH_X_TAGS` should be first and `raw_tag in valid_tags` should be in `elif`.

xxyzz:
You may want to remove some code in `extract_examples()` in file `page.py`, I didn't remove them because I want minimal changes to old code for easier review and also I can't do it...

kristian-clausal:
I took a look at the code in my editor, and the type checker noticed that `extract_template_ja_usex` and `clean_example_empty_data` don't return anything. `clean_example_empty_data` shouldn't return anything, it modifies in-place, so the signature is just wrong, but the ja_usex needs to return something because its return value is appended to examples; it also modifies in-place. Please take a look.

xxyzz:
I'll create another pr to fix this, I forget to check the "ja-x" template... Thanks!

PR #810, opened 2024-09-10
[en] return `ExampleData` from `extract_template_ja_usex()`

xxyzz:
forget to return it...

### DIFF
--- a/src/wiktextract/extractor/en/example.py
+++ b/src/wiktextract/extractor/en/example.py
@@ -86,9 +86,11 @@ def extract_quote_templates(
     ):
         roman = clean_node(wxr, None, i_tag)
         break
-    return ExampleData(
+    example_data = ExampleData(
         text=text, ref=ref, english=translation, roman=roman, type="quote"
     )
+    clean_example_empty_data(example_data)
+    return example_data
 
 
 def extract_template_ja_usex(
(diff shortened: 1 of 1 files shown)


src/wiktextract/extractor/en/example.py
```diff
@@ -118,6 +120,8 @@ def extract_template_ja_usex(
     example_data["literal_meaning"] = clean_node(
         wxr, None, node.template_parameters.get("lit", "")
     )
+    clean_example_empty_data(example_data)
+    return example_data
```
kristian-clausal, line 124: `example_data` seems to still be the same object as `parent_data`; I think this should be copied. Same in `extract_template_zh_x`, you're receiving `parent_data`, modifying it and returning it. I *think* this results in the function returning the same object over and over again, so the objects in `examples` are the same object.

kristian-clausal (changes requested):

src/wiktextract/extractor/en/example.py
xxyzz: No need to copy at here. The difference is "zh-x" create two examples data, so it need to copy the example data from its parent list before adding new data. But "ja-x" only create one data, and the parent list data is not added.

I have tested some Japanese page, code runs fine.

kristian-clausal (approved):

PR #820, opened 2024-09-16
Failing tests for reversing parse with clean_node

kristian-clausal:
When parsing lists, clean_node() will not reverse the transformation correctly:

```
# line 1
## line 2
#: example 1
#: example 2
```

becomes:

```
# line 1
## line 2
#:#: example 1
#: example 2
```

Currently only have a failing test, I'm making this pr just so that I remember to take a look later. The parse tree seems correct.

I also moved a `clean_node()` tests from `test_clean.py` into `test_clean_node.py`

### DIFF
--- a/tests/test_clean.py
+++ b/tests/test_clean.py
@@ -324,40 +324,6 @@ class WiktExtractTests(unittest.TestCase):
             "some text some other text",
         )
 
-    def test_bold_node_in_link(self):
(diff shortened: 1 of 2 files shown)


kristian-clausal:
It's in Wtp.to_wikitext or deeper, so closing this PR and making something in wikitextprocessor instead.

PR #823, opened 2024-09-17
[en] extract "zh-x" example template in etymology section

xxyzz:
Commonly used in Chinese Chengyu pages, extract them to the new "etymology_examples" field, same as zh edition.

Example pages: [一毛不拔](https://en.wiktionary.org/wiki/一毛不拔), [焚膏繼晷](https://en.wiktionary.org/wiki/焚膏繼晷)

zh edition code also have some examples in tests.


xxyzz:
But it will need more changes... Maybe we could do that if we find other similar use cases.

I haven't use these function arguments of the `clean_node()` much in non-en edition code, I usually filter the nodes before calling this function. I kind feel these functions are not flexible enough, I'd like to find a node under certain node to avoid finding some unwanted nodes.

PR #824, opened 2024-09-17
Simple english

kristian-clausal:
This is meant to be an "introductory" extractor, because of how small and reasonably consistent Simple English Wiktionary is. It has a simpler structure than usual, because there is not need to handle any language other than English.


kristian-clausal:
To be done: tests, tags, all sorts of things.

kristian-clausal:
I've copy-pasted some code from other extractors by @xxyzz, but mostly just Pydantic models.

src/wiktextract/extractor/simple/preprocess.py
```diff
@@ -0,0 +1,113 @@
...
+from .text_utils import (
+    HEADING_RE,
+    POS_STARTS_RE,
+    POS_TEMPLATE_NAMES,
+    SECTION_STARTERS,
+    TEMPLATE_NAME_RE,
+)
+
+
+def preprocess_text(text: str, page_title="") -> str:
```
xxyzz, line 12: IMO code here changes wikitext is too complicated and risky, it needs at least some tests to let others (and yourself a few weeks later) to know what's it does...

I'm not fully aware the page layout this function wants to change, but I think the code will be easier to maintain if this process step could be avoided.

I guess the problem is section "Word parts" and "Pronunciation" uses level 3 node and  placed above the POS level 2 node? How about extract the level 3 section data to the base `WordEntry` data, then it will be shared to all following POS sections. For example:

```python
def parse_page(
    wxr: WiktextractContext, page_title: str, page_text: str
) -> list[dict[str, Any]]:
    wxr.wtp.start_page(page_title)
    tree = wxr.wtp.parse(page_text)
    page_data = []
    base_data = WordEntry()
    for level_node in tree.find_child(LEVEL_KIND_FLAGS):
        title_text = clean_node(wxr, None, level_node.largs)
        if title_text is pos:
            new_data = base_data.model_copy(deep=True)
            extract_pos(wxr, level_node, new_data)
            page_data.append(new_data)
        elif title_text == "Word parts":
            extract_word_parts(wxr, level_node, base_data)
        elif title_text == "Pronunciation":
            extract_pronunciation(wxr, level_node, base_data)
    return [m.model_dump(exclude_defaults=True) for m in page_data]
```

The current `parse_page()` could be simplified IMO.

src/wiktextract/extractor/simple/page_utils.py
```diff
@@ -0,0 +1,50 @@
...
+    # level kinds
+    return node.children[:i], node.children[i:]  # type: ignore[return-value]
+
+
+def recurse_base_data_sections(
+    wxr: WiktextractContext, node: WikiNode, target_data: WordEntry
+) -> None:
+    # Parse a section recursively; a heading node with contents and possible
+    # subsections (always following the non-subsection content)
+    content, subsections = separate_subsections(node)
```
xxyzz, line 30: Why not use the parsed node hierarchy directly:

```python
title_text = clean_node(wxr, None, level_node.largs).lower()
if title_text.startswith(("etymology", "word parts")):
    pass
elif title_text.startswith("..."):
    pass
for next_level in level_node.find_child(LEVEL_KIND_FLAGS):
    recurse_base_data_sections(wxr, next_level, target_data)

```

xxyzz (commented):
IMO the some parts could be simplified, especially if the goal is to create a documentation and example for new editions.

And a small suggestion, I'd recommend create one pr for one wikitext section's code, and also add tests in the same implementation commit this helps me many times to understand my own code...

src/wiktextract/extractor/simple/pronunciation.py
```diff
@@ -0,0 +1,284 @@
...
+    # across a template, before expanding the template it calls whatever was
+    # passed to template_fn=. The handler function can return a string, which is
+    # inserted into the returned text, or None in case nothing special will be
+    # done with the template and it will use the normal expanded value.
+    # post_template_fn= is the same, except it happens after the template has
+    # already been expanded and receives that text as a parameter. If you want
+    # to replace templates based on their names or arguments before expansion,
+    # use template_fn=, if you want access to the expanded text without doing
+    # it manually, use post_template_fn=
+    def parse_pronunciation_template_fn(
```
xxyzz, line 59: I'm a bit disagree with this `clean_node()` template process functions practice. Normally we need the parsed expanded template node structure and its parameters.

```python
expanded_node = wxr.wtp.parse(
    wxr.wtp.node_to_wikitext(template_node), expand_all=True
)
```

and also we could use [`TemplateNode.template_parameters()`](https://github.com/tatuylonen/wikitextprocessor/blob/272742c89ec9a82208022d5f6063a1736c90b7fb/src/wikitextprocessor/parser.py#L569) method if use template args is easier.

I'd recommend create function for each template, create extracted `Sound` data directly in these functions and avoid processing plain text and long nested functions.

src/wiktextract/extractor/simple/page.py
```diff
@@ -0,0 +1,202 @@
...
+
+def parse_page(
+    wxr: WiktextractContext, page_title: str, page_text: str
+) -> list[dict[str, Any]]:
+    """Parse Simple Wiktionary page."""
+    # Unlike other wiktionaries, Simple Wikt. has a limited scope: only English
+    # words. The pages also tend to be much shorter and simpler in structure,
+    # which makes parsing them much simpler.
+
+    if page_title.startswith("Appendix:") or page_title == "Main Page":
```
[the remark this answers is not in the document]
kristian-clausal: I added that `.startswith("Appendix:")` later in the process, so it was a problem. I probably just have an old .db file.

src/wiktextract/extractor/simple/preprocess.py
kristian-clausal: In retrospect this might be true, I'll take a look at it. You start of solving one problem, and then things escalate... There were many steps that led to what you see in the code.

src/wiktextract/extractor/simple/page_utils.py
kristian-clausal: That's remnants of when "contents" was/is useful as a clump. If you want to run it through clean_node() whole, then you can just do `clean_node(..., contents, ...)`. I'll clean it up.

src/wiktextract/extractor/simple/pronunciation.py
kristian-clausal: I think it's because I am so influenced by the messiness of English Wiktionary source wikitext. The main English Wiktionary extractor has to be a complex mess because the source is a complex mess, and you are correct that other extractors can get away with much simpler code. I'll take a look at how it's done on Simple Wiktionary and whether you can trust the source to be "stable".

xxyzz: I totally understand and agree with you the en code situation...

I think we're probably fine with processing expanded wikietxt and HTML nodes, even if the template changed dramatically it's still manageable, at least couldn't be worse then processing plain text.

src/wiktextract/extractor/simple/page_utils.py
kristian-clausal: Looking at this, I will keep things as is. There is nothing harmful here, and as the processing functions are they take a whole `contents` block as is, which is needed because we want a contents-wide context.

xxyzz: Usually I'd like to pass the level node to each section's extract function, because I could then use `WikiNode.find_child()` on this node. If I want to convert the text of this section but exclude subsection I could pass `WikiNode.invert_find_child(LEVEL_KIND_FLAGS)` to `clean_node()`.

kristian-clausal: That sounds good. I've removed the preprocessing (except for removing `text block start` template stuff), so now the main page loop is going through each level in order.

One problem with this approach is that *sometimes* you have a situation like this:

```
== Noun ==
...
=== Pronunciation ===
...
=== Word parts ===
...
```

at the end of an article, so the pronunciation and word parts stuff is meant to go with the Noun, instead of preceding it (as is the usual way). This is why I ended up with the preprocessing stuff earlier.

xxyzz: I have the similar problem in zh edition, I use a recursive [`parse_section()`](https://github.com/tatuylonen/wiktextract/blob/3218b933c35638294b6886a4bdb0d94e2bcc3788/src/wiktextract/extractor/zh/page.py#L67) function processing any level node at any place. It adds the extracted data to the last `WordEntry` if we already have some POS section data or adds to `base_data`. Some sections probably do more than that, but I think it's the general idea.

kristian-clausal: I solved this by changing from iterators to lists and also parsing heading titles before-hand. That way, each time parsing a pron/etym section, you can check if any of the following titles are POS headings and if not, use the last POS data instead of putting stuff in `base_data`.

src/wiktextract/extractor/simple/models.py
```diff
@@ -0,0 +1,110 @@
...
+    model_config = ConfigDict(title="Simple English Wiktionary")
+
+    word: str = Field(description="Word string")
+    # For Simple English, the language is always English
+    forms: list[Form] = Field(default=[], description="Inflection forms list")
+    lang_code: str = Field(default="", description="Wiktionary language code")
+    lang: str = Field(default="", description="Localized language name")
+    pos: str = Field(default="", desription="Part of speech type")
+    pos_title: str = ""
+    pos_num: Optional[int] = None
```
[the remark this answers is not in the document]
kristian-clausal: Changed to `-1" as the default.

src/wiktextract/extractor/simple/page.py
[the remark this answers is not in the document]
kristian-clausal: I recreated the .db file and `Appendix:` are present:

```
2024-09-18 12:44:29,205 ERROR: ///////////////      page_title='Appendix:Roman numerals'
2024-09-18 12:44:29,716 INFO:   ... 28674/48710 pages (58.9%) processed, 00:00:07 remaining
2024-09-18 12:44:30,716 INFO:   ... 32194/48710 pages (66.1%) processed, 00:00:06 remaining
2024-09-18 12:44:31,039 DEBUG: 've: A cell that's not exactly 2 lines: 'Negative'
2024-09-18 12:44:31,043 DEBUG: 've: A cell that's not exactly 2 lines: "haven't"
2024-09-18 12:44:31,043 DEBUG: 've: A cell that's not exactly 2 lines: "hasn't"
2024-09-18 12:44:31,043 DEBUG: 've: A cell that's not exactly 2 lines: "hadn't"
2024-09-18 12:44:31,044 DEBUG: 've: A cell that's not exactly 2 lines: 'None'
2024-09-18 12:44:31,716 INFO:   ... 35988/48710 pages (73.9%) processed, 00:00:04 remaining
2024-09-18 12:44:32,716 INFO:   ... 39932/48710 pages (82.0%) processed, 00:00:03 remaining
2024-09-18 12:44:33,755 INFO:   ... 42897/48710 pages (88.1%) processed, 00:00:02 remaining
2024-09-18 12:44:33,806 ERROR: ///////////////      page_title='Appendix:English numerals'
```

Not a high priority, just a bit weird.

src/wiktextract/extractor/simple/page.py
```diff
@@ -0,0 +1,202 @@
...
+
+    #####
+
+    # Parse the page, pre-expanding those templates that are likely to
+    # influence parsing; these are problematic templates that generate stuff
+    # that "should" be present for the page to parse better
+    tree = wxr.wtp.parse(
+        page_text,
+        pre_expand=True,
+        additional_expand=ADDITIONAL_EXPAND_TEMPLATES,
```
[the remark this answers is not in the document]
kristian-clausal: Yeah, prematurely added ADITIONAL_EXPAND_TEMPLATES and the config doesn't enable analyzing templates.

src/wiktextract/extractor/simple/page.py
```diff
@@ -0,0 +1,149 @@
...
+
+    flattened_levels = list(tree.find_child_recursively(LEVEL_KIND_FLAGS))
+    headings_of_levels = [
+        clean_node(wxr, None, level.largs[0]).lower()
+        for level in flattened_levels
+    ]
+
+    for i, (heading_title, level) in enumerate(
+        zip(headings_of_levels, flattened_levels)
+    ):
```
xxyzz, line 99: Could be further simplified, please see my explanation in the next comment. 

```suggestion
    for level_node in tree.find_child_recursively(LEVEL_KIND_FLAGS):
        heading_title = clean_node(wxr, None, level_node.largs).lower()
```

and my suggestion code could have bugs, don't apply them directly...

src/wiktextract/extractor/simple/page.py
```diff
@@ -0,0 +1,149 @@
...
+        # be anything there. Previous version of this code looked through that
+        # stuff and would spit out warnings.
+
+        # Collect Etymology (Word parts), Pronunciation data.
+        # Ignore Description for now. XXX maybe implement something if relevant.
+
+        # print(f"=== {heading_title=}")
+
+        if m := POS_STARTS_RE.match(heading_title):
+            pos_data = base_data.copy(deep=True)
```
xxyzz, line 110: I'd recommend avoid building long regex pattern if possible.

```suggestion
        m = re.search(r"\s*\d+$", heading_title)
        if m is not None:
            heading_title = heading_title[:m.start()]
            pos_num = int(m.group().strip())
        if heading_title in POS_DICTIONARY:
            new_data = process_pos(wxr, base_data.model_copy(deep=True))
            word_datas.append(new_data)
        elif heading_title == "pronunciation":
            process_pron(wxr, level, word_datas[-1] if len(word_datas) > 0 else base_data)
```

New `WordEntry` is only added to `word_datas` if there was a level 2 POS section, and because level 3 can't be the child of next level 2 node, we don't need to check afterward titles.

I think any other cases will all mean the original wikitext are very wrong and we don't need to handle a corrupted layout.

src/wiktextract/extractor/simple/page.py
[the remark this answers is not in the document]
xxyzz: I didn't know simple Wiktionary doesn't have "Appendix" namesapce and they put these pages in the main article(namespace id is 0) namesapce...

src/wiktextract/extractor/simple/preprocess.py
```diff
@@ -0,0 +1,12 @@
+import re
+
+TEXT_BOX_RE = re.compile(r"(?m)\n?{{text box start[^\n$]+{{text box end}}\n?")
```
xxyzz, line 3: Only [one page](https://simple.wiktionary.org/wiki/poem) uses this template and it's doesn't interfere with extracting gloss lists, because its content uses ":" indent list not `#` numbered gloss list. I think this file could be removed now.

src/wiktextract/extractor/simple/page.py
kristian-clausal: This doesn't work because the layout is more complicated.

src/wiktextract/extractor/simple/page.py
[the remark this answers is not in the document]
kristian-clausal: Oh, that would explain it... :face_exhaling:

src/wiktextract/extractor/simple/page.py
xxyzz: Are there really exists such layout? That will render PronB section under the NounA section. Why would they do that, that doesn't make sense...

kristian-clausal: ... I can't find any pages where it happens, except a couple of ones where a section is named wrongly (like in `valiant`, which I've already corrected on Wiktionary). Yet again, I've seen ghosts. :expressionless:

kristian-clausal: By "it happens" I mean that NounC and PronC are that way around, the PronA and PronB stuff is the standard way of doing it on Simple English Wiktionary!

xxyzz: The code also works for the non-standard PronC under NounC layout, don't worry too much about rare cases...

I think we'll both agree we can't really expect extract much data from such a small dump file, and its most pages don't have more than two POS sections. I guess simpler and cleaner code as a demonstration and documentation is more important for this edition?

kristian-clausal: The NounC and PronC isn't what happens, only PronA and PronB happens. I've been messing up explaining this.

src/wiktextract/extractor/simple/page.py
```diff
@@ -0,0 +1,125 @@
...
+    # Unlike other wiktionaries, Simple Wikt. has a limited scope: only English
+    # words. The pages also tend to be much shorter and simpler in structure,
+    # which makes parsing them much simpler.
+    # https://simple.wiktionary.org/wiki/Wiktionary:Entry_layout_explained
+
+    if page_title in (
+        "Main Page",
+        "Main page",
+    ) or page_title.startswith("Appendix:"):
+        return []
```
xxyzz, line 29:
```suggestion
    if page_title == "Main Page" or page_title.startswith("Appendix:"):
        return []
```

redirect pages are filtered in `page_handler()`: https://github.com/tatuylonen/wiktextract/blob/a4f804451f4fe1216067c423bdc53a43ee7f38de/src/wiktextract/wiktionary.py#L55

`wxr.wtp.start_page()` is also called in that function.

src/wiktextract/extractor/simple/page.py
```diff
@@ -0,0 +1,125 @@
...
+        # be anything there. Previous version of this code looked through that
+        # stuff and would spit out warnings.
+
+        # .find_child_recursively() (in contrast with find_child()) yields
+        # a flattened tree, not just direct children of root
+
+        # print(f"=== {heading_title=}")
+        heading_title = clean_node(wxr, None, level.largs[0]).lower()
+
+        if m := re.search(r"\s+\d+h$", heading_title):
```
xxyzz, line 97: typo?
```suggestion
        if m := re.search(r"\s+\d+$", heading_title):
```

src/wiktextract/extractor/simple/page.py
```diff
@@ -0,0 +1,125 @@
...
+    # pronunciation data and etymological data is collected and updated in this
+    # base_data object that can be copied for more specific entries.
+    # Simple English Wiktionary has these sections under Level 3 ("===")
+    # headers, which put them in an awkward hierarchy with the main page and
+    # Part-of-Speech sections, which are higher; they usually appear before
+    # POS sections, but a few pages have a more complex structure. This
+    # is handled by simply flattening the parse tree later, and handling
+    # sections in a linear order.
+    base_data = WordEntry(
+        word=wxr.wtp.title or "ERROR_NO_TITLE",
```
xxyzz, line 77:
```suggestion
        word=page_title,
```

`wxr.wtp.title` can't be `None`, use `str` could shut up type checking tools

src/wiktextract/extractor/simple/section_titles.py
```diff
@@ -0,0 +1,114 @@
...
+    "acronym & initialism": "acronym",
+}
+
+# POS aliases; use this to figure out if a heading is a POS heading
+POS_HEADINGS = { k: v for k, v in POS_DATA.items() }
+
+for k, v in POS_HEADINGS_MAP.items():
+    if k in POS_HEADINGS or v not in POS_DATA:
+        continue
+    POS_HEADINGS[k] = POS_DATA[v]
```
xxyzz, line 114: If only have one title needs to map, better add it directly to `POS_DATA`.

`POS_DATA["acronym & initialism"] = POS_DATA["acronym"]`

or

```python
for k, v in POS_HEADINGS_MAP.items():
    POS_DATA[k] = POS_DATA[v]
```

no need to duplicate whole dictionary for a few keys.

src/wiktextract/extractor/simple/text_utils.py
```diff
@@ -0,0 +1,17 @@
+import re
+
+from .section_titles import POS_DATA
+
+# A regex for detecting a heading that starts like a POS: "Noun 1"
+POS_STARTS_RE = re.compile(
```
xxyzz, line 6: I think this long regex could also be removed.

IMO the code extract other sections like gloss lists and sounds have more room to be further simplified, still fell too complicated for the simple layout used in this edition...

xxyzz (commented):
Just some minor suggestions.

src/wiktextract/extractor/simple/page.py
kristian-clausal: Oh crap. Yeah, that's a typo... I'm using a modal editor at the moment with heavily rebound keys (HJKL is a terrible historical accident, ask me about it), "h" is my "insert text" key.

src/wiktextract/extractor/simple/section_titles.py
kristian-clausal: Aaah, of course. I got stuck thinking about if it would be possible to do internal references like that when first writing out the dictionary literal, but somehow I never figured this.

src/wiktextract/extractor/simple/page.py
xxyzz: I used to use vim, but I'm too lazy to customize any keys even now... I vaguely remember that's cursor movement keys, and "i" for insert?

src/wiktextract/extractor/simple/text_utils.py
kristian-clausal: I've removed it and replaced it with a "detect last digits" regex and checking the start against POS_HEADINGS, I'm just working on some stuff involving it so I'll commit later.

pyproject.toml
```diff
@@ -30,7 +30,7 @@ classifiers = [
 ]
 requires-python = ">=3.9"
 dependencies = [
-    "importlib_resources; python_version < '3.10'",
```
xxyzz, line 33: Should remove this line if min-version is set to 3.10, the test action fail before is because the Python 3.9 test matrix is not removed: https://github.com/tatuylonen/wiktextract/blob/1e6ea054d3ae18a1c4fc688356675072c4f259f6/.github/workflows/test.yml#L30

kristian-clausal: Yeah, sorry, I tried out stuff earlier and completely forgot these...

kristian-clausal: I reverted this branch to 3.9, switching to 3.10 should be a separate operation.

PR #829, opened 2024-09-20
[en] move some en edition code to "extractor/en" folder

xxyzz:
- should reduce load time and memory usage for non-en code
- rename some en test files with "test_en_*" prefix
- fix a conflict function name in "test_en_pronunciation.py"
- "type_utils.py" is also moved cuz it's mainly used by en code and can't represent non-en pydantic models
- some imports are removed from "__init__.py" because they are en edition code therefore inappropriate to be added to package root but still can be imported from new paths

Mostly boring imports changes, non-en code now load much faster(reduce about 3.5s).


src/wiktextract/__init__.py
```diff
@@ -21,12 +18,8 @@
...
     "parse_wiktionary",
     "reprocess_wiktionary",
-    "PARTS_OF_SPEECH",
     "parse_page",
     "extract_thesaurus_data",
     "extract_namespace",
     "extract_categories",
-    "sort_tags",
-    "tag_categories",
-    "valid_tags",
```
kristian-clausal, line 31: valid_tags is the authorative source of what tags are valid in all extractors, and shouldn't be moved. This is one of the things that should remain from tags.py at the top level, there are the allowed output tags.

kristian-clausal: Same for tag_categories, etc. these are not meant to be En-exclusive.

src/wiktextract/extractor/en/page.py
```diff
@@ -20,56 +20,48 @@
...
-from wiktextract.wxr_logging import logger
-
+from ...wxr_context import WiktextractContext
+from ...wxr_logging import logger
 from ..ruby import extract_ruby, parse_ruby
 from ..share import strip_nodes
 from .example import extract_example_list_item, extract_template_zh_x
+from .form_descriptions import (
+    classify_desc,
+    decode_tags,
```
kristian-clausal, line 51: We should make a generic version of `decode_tags` at some point (at the moment it's couple too much with en stuff), `classify_desc` unfortunately can't be made generic.

`distw` could be at the top level, it just tells the Levenshtein distance between two words (how similar they are), so should be applicable to anything that the `Levenshtein` package supports, which should be at minimum be most languages with Latin script or maybe all scripts.

kristian-clausal (changes requested):

src/wiktextract/extractor/en/page.py
kristian-clausal: This is not very important, just for the future.

src/wiktextract/__init__.py
xxyzz: But other editions now have their own `tags.py` files and have much smaller data and also new tags. I think import from each edition's `tags.py` is more appropriate?

src/wiktextract/extractor/en/page.py
xxyzz: I don't know if we really want a complex function for all editions...

src/wiktextract/__init__.py
kristian-clausal: valid_tags is not en edition data. It's a dict of all allowed tags (besides upper-case tags for placenames). This is the list that says "infinitive" is an ok tag. This is basically the terminology of linguistics, and should be applicable to every language edition, because you are supposed to translate tags so that they are valid_tags. It's the specification for what tags are allowed to contain.

kristian-clausal: The only reason it is used in the English edition as a source for directly checking tags in text is because it's an English language list and "nominative" in English text will be translated into a "nominative" tag. If it was the Finnish edition and it came across "nominatiivi", it should still end up as "nominative" through a translation function or map, but the target is that the end result should be in `valid_tags`. If some language edition has a term that could fit into valid_tags (and can be translated to English linguistic terminology), then it should be added to valid_tags. The exception here is place-names and dialect/language names, "upper-case" terms that are so messy they aren't part of valid_tags (and should also be translated).

xxyzz: But there is not really need to load a huge file even if we're not using it... Move the file doesn't prevent users to import it.

I really don't see the purpose of loading a huge file in `__init,py__`, if you insist doing this, I'd suggest we move that dictionary or just the keys to the root folder.

kristian-clausal: Yes, move the dict, that is what I asked you to do. As long as `valid_tags` and the things is relies on are not the En extractor.

xxyzz: I could move it but I still a bit disagree. Could you please explain why `valid_tags` is relied? It's clear only used in en code. Why we have to load a huge file for all code even if it's not used?

xxyzz: Sigh... I already created `tags.py` for each edition, most these data are translated manually to match en tags and some(very few) are not existed in en tags, I think you'd understand every edition would have some unique data.

xxyzz: Sign.. I don't know if you really understand how time consuming I translate these files...

And if we really need to import that huge unnecessary data, we don't need to add it to `__init__.py`, we could use `from wikitextract.extractor.en.tags import valid_tags` to load it later.

kristian-clausal: If you're worried about the impact of `valid_tags` in __init__.py, then we can make it a manual import.

xxyzz: My current code already only import "en.tags" when it's actually used...

And about add unique tags only used in certain edition to it, I really don't want do that and I don't see the point....

kristian-clausal: If your tags are good and proper *linguistic tags* that contain information about a specific language, that doesn't change just because the edition changes: if a word is described as something in one edition, that description should be possible in other editions. If they're tags that don't belong in general circulation, then you should consider if they should be tags at all. Do you have any examples? I am looking at some of these, and you have the tag "officialese" in the German tags; there is a similar tag "bureaucratese" in valid_tags, and the En mapping has "officialese" -> "bureaucratese". Later on the page I see a couple of untranslated terms: "familiär" is most probably the tag "familiar" and "gehoben" should most probably be "formal". But 99.9% seems like stuff in `valid_tags` or uppercase terms that refer to places and dialects/languages.

xxyzz: I think you need to change your position from en maintainer to non-en maintainer... non-en code tags are manually translated, of course I can't guaranty every tag is translated perfectly. Now every edition has it's own tags.py file and they are manually translated at best efforts to match en tags. I really don't want keep every one of them synced 100%... Can you understand my situation? And because they are created manually, I don't need to import en.tags.

kristian-clausal: At a quick look, I didn't see anything weird with your tag translations. 99% of the time the tag is something that is already in `valid_tags`, or it's a word like "jocular" that has a synonym of it in `valid_tags` ("humorous"), or it's an upper-case tag referring to a place or dialect. There really seems to be no real conflict here, all of this should be easy to just check by validating the values in your translation maps against `valid_tags` (ignoring uppercase tags). AFAICT you've been following the contents of `valid_tags`. 99% of your non-uppercase tags are already in `valid_tags`.

kristian-clausal: Even if you don't need to import `valid_tags` (which is not en.tags!), it should still be at the top level.

kristian-clausal: I don't understand what you are so terrified of. You've already done correctly, the tags need to be translated, they just need to conform to `valid_tags`. There is really nothing much else to do here except keep `valid_tags` at the top level as an authoritative source (which can be amended with new tags when needed) for what tags are valid in the JSON output.

xxyzz: I think what you want is a list of all currently used tags combined in one place? I think that could be generated automatically like these JSON schema files and upload it to GitHub Page.

IMO we don't need this list in our code because no-valid tags can't be translated.

kristian-clausal: I AM PRAISING THAT YOUR TAGS ARE GOOD.

xxyzz: You probably don't understand my irrational fear about these tags.py files... sorry about that...

So how about the solution I mentioned above?

kristian-clausal: Sorry about shouting... I will ask Tatu about it next week, but I suspect he will say that we are going to keep a "central authority", that's the point of making these tags, so that the data is interchangeable.

xxyzz: Let me summary the situation... I understand you want all approved tags saved at one place, what I don't agree is saving it in a large Python file and load it in all edition code. As I explained, each non-en edition tags are translated from a dictionary manually created so we don't need this huge file to verify translated tags because all tags in there are already valid(with a few outliers).

I could remove these outliers but I'm scared of adding new tags.

xxyzz:
I have moved "tags.py" and "topics.py" back to the "src/wiktextract" folder.

kristian-clausal:
We can move stuff around later. :+1:

kristian-clausal (approved):

kristian-clausal:
Tatu said that `valid_tags` and other such should be kept for validation of tags between editions. All editions should, as far as we can make it possible, have the same standard tags. `tags.py` has some stuff that belongs at top level (`valid_tags`) and some that don't (all the maps of English strings -> tags).

PR #843, opened 2024-10-02
clean_value: Remove whitespace before Category links

kristian-clausal:
Can't find any easy documentation, but wikitext behavior seems to be that if there is any whitespace, including newlines, before a [[Category:...]] link that is not a "displayed" [[:Category:...]] link, all whitespace before that is stripped away. So:

```
foo

[[Category:test]]bar
```
becomes `foobar`.

### DIFF
--- a/tests/test_clean.py
+++ b/tests/test_clean.py
@@ -8,7 +8,7 @@ from wiktextract.thesaurus import close_thesaurus_db
 from wiktextract.wxr_context import WiktextractContext
 
 
-class WiktExtractTests(unittest.TestCase):
+class CleanTests(unittest.TestCase):
     def setUp(self) -> None:
         self.wxr = WiktextractContext(Wtp(), WiktionaryConfig())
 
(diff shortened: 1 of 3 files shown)


kristian-clausal:
A minimal change to a regex in clean_value to fix an issue with some templates generating unexpected newlines that were not appearing in website output on Wiktionary. See more in Wikitextprocessor issue 314.

tests/test_clean.py
```diff
@@ -8,9 +8,20 @@
 from wiktextract.wxr_context import WiktextractContext
 
 
-class WiktExtractTests(unittest.TestCase):
+class CleanTests(unittest.TestCase):
     def setUp(self) -> None:
         self.wxr = WiktextractContext(Wtp(), WiktionaryConfig())
+        self.NAMESPACE_DATA = {
```
xxyzz, line 14: Probably don't need this? By default, `self.wxr.wtp` should load the en edition namespace data.

kristian-clausal: I thought I had a problem, but that might have been because I was trying out git worktrees and it turned out that nothing worked because my venv would (of course) still point to the original editable repo and not the worktree copy-branch.

kristian-clausal: Yeah, it was worktrees, and then I just "fixed" an issue that wasn't an issue.

PR #847, opened 2024-10-03
[en] move analyze template code from wikitextprocessor

xxyzz:
part of https://github.com/tatuylonen/wikitextprocessor/issues/316

the default `WiktionaryConfig.save_ns_names` and
`WiktionaryConfig.extract_ns_names` values are set to more general namespaces, previous settings for en edition are moved to the new `data/en/config.json` config file.

Require tatuylonen/wikitextprocessor/pull/318

I have tested the code on en edition dump file, the extracted JSON data seem fine.


src/wiktextract/import_utils.py
```diff
@@ -5,14 +5,11 @@
 
 def import_extractor_module(
     lang_code: str, module_name: str
-) -> types.ModuleType:
-    default_module_name = f"wiktextract.extractor.en.{module_name}"
+) -> types.ModuleType | None:
```
kristian-clausal, line 8: I've added --edition and --dump-file-lang-code to the regen script because they were missing for the main En invocation, hopefully things work out.

src/wiktextract/extractor/en/analyze_template.py
```diff
@@ -0,0 +1,138 @@
...
+            tag_cnts[tagname] -= 1
+        elif not end_slash:
+            tag_cnts[tagname] += 1
+    contains_unbalanced_html = any(v != 0 for v in tag_cnts.values())
+    # if contains_unbalanced_html:
+    #     print(name, "UNBALANCED HTML")
+    #     for k, v in tag_cnts.items():
+    #         if v != 0:
+    #             print("  {} {}".format(v, k))
+
```
kristian-clausal, line 132: The original had some code for recursively marking a template to be expanded if any of its child templates needed to be expanded, how does that work here?

kristian-clausal (changes requested):

src/wiktextract/extractor/en/analyze_template.py
```diff
@@ -0,0 +1,138 @@
...
+    return (
+        contains_list
+        or contains_unpaired_table
+        or contains_table_element
+        or contains_unbalanced_html
+    )
```
kristian-clausal, line 138: Instead of `or`ing here, I think this could all be technically done with early returns.

src/wiktextract/extractor/en/analyze_template.py
xxyzz: pre-expanded template will set `expand_recurse()`'s `expand_all` arg to `True`: that change is in tatuylonen/wikitextprocessor/pull/318 file "core.py" line 1464

src/wiktextract/extractor/en/analyze_template.py
xxyzz: I didn't think about that... I mostly want minimal changes to these complex code for easier compare.

src/wiktextract/extractor/en/analyze_template.py
kristian-clausal: AFAICT, that's top down, but the the purpose of the original code was to "bubble up". from the bottom. If a template calls a template that calls a template that should be pre-expanded, the old code would (in the later loop) also mark the high templates as pre-expanded. At the moment we do not seem to do that. If a template calls a template with `need_pre_expand`, it also was marked with `need_pre_expand`.

src/wiktextract/import_utils.py
xxyzz: The default en module is removed at here is because I think load en edition code for not implemented edition is inappropriate, return `None` and raise exception in the caller code is fine.

kristian-clausal: Yeah, it would have just broken kaikki's regen for a while if I hadn't thought about it. Should be fine.

src/wiktextract/extractor/en/analyze_template.py
xxyzz: I'll restore these code, really hard to change them...

kristian-clausal (approved):

kristian-clausal:
Thanks for doing this, making these changes can be a bit scary...

xxyzz:
I think it's kind alright? Not much code really changed.

PR #859, opened 2024-10-09
[ko] extract note list and linkage list

xxyzz:
terrible wikitext layout


kristian-clausal:
> terrible wikitext layout

I think we can pretty much copy paste this for every commit we make.

xxyzz:
Not really... fr, pl and some other editions are quite good at maintaining a standard layout for most pages, but some smaller editions like ja and ko are really terrible at this.

PR #871, opened 2024-10-15
[en] Add 'keine' to conditionally ignore cells for German

kristian-clausal:


kristian-clausal:
E-mailed issue, see Beamter / German.

PR #888, opened 2024-10-28
[en] silence invalid uppercase tag check JSON debug messages

xxyzz:
too many this kind of messages bury the real check JSON messages

`valid_tags` and `uppercase_tags` global variables should use uppercase, but en code import them in many places and website code could also be affected so I didn't rename them.

### DIFF
--- a/src/wiktextract/wiktionary.py
+++ b/src/wiktextract/wiktionary.py
@@ -245,17 +245,16 @@ def check_tags(
         # non-English editions).  Tag values should be standardized across
         # editions, except for uppercase tags (e.g., regional variants).
         if wxr.wtp.lang_code in ("en",):  # Check edition
-            from .tags import valid_tags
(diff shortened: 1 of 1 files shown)


kristian-clausal:
This actually doesn't fix the issue. uppercase_tags doesn't have "Egyptian-Arabic", because that tag is generated and thrown into the tags even though it's not in uppercase_tags. This actually needs two different error messages, the other can have all normal "invalid tags" messages, the other can have "invalid tag but it's uppercase" tags; that should clean it up a bit.

xxyzz:
Err, what? I didn't know that. At least the data in the list won't create debug messages, maybe this will remove most of them. But if there are still many combined uppercase tags, other JSON errors like "empty string" will still be difficult to find.

kristian-clausal:
I've made a PR to separate the error messages, check it out to see if I did any embarrassing mistakes.

PR #889, opened 2024-10-28
[en] Split tag validation error into two

kristian-clausal:


xxyzz (approved):
A small typo in the first message: `f"invalid uppercase tag {tag} not in uppercase_tags".`

The second message probably should only say it's not in `valid_tags`?

but kaikk.org's check JSON page doesn't combine similar messages because the full JSON string in the message, change the message probably still hard to find other non-tag messages.

kristian-clausal:
Ah, I assumed this just passed the debug messages somewhere. The second message still checks `uppercase_tags`, so it should be correct.

kristian-clausal:
By changing the `called_from` we can filter away the uppercase tag message because  the kaikki.org page will only pull errors with the default `called_from` string. Didn't need to change anything on the kaikki side.

xxyzz:
Thanks!

PR #898, opened 2024-11-04
Fix filenames

timendum:
Hi,
two files ended with a space (' '), this cause some problems on Windows and i think it's not intended.


xxyzz:
Thanks!

kristian-clausal:
Oh wow, you're right. I had to `ls -Q` to even see that. Thanks for the PR!

PR #899, opened 2024-11-04
Fix README.MD typo

Alex-Programs:
Changed http:// to https:// and removed the space halfway through the link so that it renders properly.


xxyzz:
Thanks!

Issue #906, opened 2024-11-11
Term labels sometimes mistaken as forms


Ian-07:
For some pages that use the {{term-label}} template, wiktextract is incorrectly adding some of those labels to the list of forms.

From [vai](https://en.wiktionary.org/wiki/vai) on Wiktionary:

> vai (plural [vais](https://en.wiktionary.org/wiki/vais#English)) ([Bangladesh](https://en.wikipedia.org/wiki/Bangladeshi_English), [West Bengal](https://en.wikipedia.org/wiki/West_Bengal), chiefly [Benglish](https://en.wikipedia.org/wiki/Benglish))

From [vai](https://kaikki.org/dictionary/All%20languages%20combined/meaning/v/va/vai.html) on kaikki.org:

```
  "forms": [
    {
      "form": "vais",
      "tags": [
        "plural"
      ]
    },
    {
      "form": "Benglish",
      "tags": [
        "West-Bengal"
      ]
    }
  ],
```

kristian-clausal:
Happily, this was a simple fix; wiktextract didn't recognise 'Benglish' as a valid taggable string, so it assumed it had to be a weird inflected form. Adding 'Benglish' to `uppercase_tags` fixes things in this specific instance.

Ian-07:
Here is another example of this which this time doesn't seem to be related to uppercase_tags: [chuunibyou](https://kaikki.org/dictionary/All%20languages%20combined/meaning/c/ch/chuunibyou.html), [Flanderization](https://kaikki.org/dictionary/All%20languages%20combined/meaning/F/Fl/Flanderization.html), and [Nuzlocke](https://kaikki.org/dictionary/All%20languages%20combined/meaning/N/Nu/Nuzlocke.html) all use the {{term-label}} template with the "fandom slang" label, and all of them erroneously have "lifestyle" as one of the forms.

kristian-clausal:
Kaikki has finally updated after bug fixes (I had the flu for a week, but there were also computer bugs *badumtsh*), please take a look if you have the time.

Ian-07:
Everything looks good now, thanks.

PR #909, opened 2024-11-13
Added "Bangladesh" and "Hinglish" to uppercase_tags

Ian-07:
see https://github.com/tatuylonen/wiktextract/issues/906


kristian-clausal:
Thank you for your contribution! These kinds of small PRs that add data are always welcome.

PR #915, opened 2024-11-21
[en] don't extract converted topic data as form

xxyzz:
#906

Not sure if this is the correct fix.


kristian-clausal:
The underlying problem here is that `Template:term-label` has a newline that forces it to be on the next line. It looks like a head-template, but isn't; there should be a form there, but there isn't.

kristian-clausal (commented):
I think this needs some special handling for Template:term-label. I will try to figure out. The bigger problem is getting the right incantations to commit to the PR, wish me luck.

kristian-clausal:
I finally figured out how to commit to other people's PRs when they're remote. :+1:

kristian-clausal:
This should hopefully do it.

PR #921, opened 2024-11-25
[en] Fix out-of-bound index error.

kristian-clausal:
Oops, indentation error... I am surprised I don't
do these more often.

Fixes a bunch of exceptions that killed kaikki extraction. For some reason, only relevant in Catalan entries, so maybe they did some bot work for those or a template changed. Not their fault, this bug is all mine.


kristian-clausal:
This was actually irrelevant, and a bug was found elsewhere.

PR #966, opened 2025-01-03
Add "--edition" option to command line examples

xxyzz:
it's not an optional argument for non-English editions


kristian-clausal:
We could make --edition obligatory.

xxyzz:
Sure, I didn't make that change because I'm afraid it could break website code.

kristian-clausal:
The regen script has explicit `--edition "en"` *and* `--dump-file-language-code "en"` so should be ok.

PR #991, opened 2025-01-16
Use `concurrent.futures.ProcessPoolExecutor`

xxyzz:
main process won't hang if one of worker process is crashed or killed

I only compared zh and th editions, just one minute less.

### DIFF
--- a/src/wiktextract/thesaurus.py
+++ b/src/wiktextract/thesaurus.py
@@ -2,15 +2,18 @@
 # merged into word linkages in later stages.
 #
 # Copyright (c) 2021 Tatu Ylonen.  See file LICENSE and https://ylonen.org
+import atexit
 import os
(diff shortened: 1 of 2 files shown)


kristian-clausal:
Let's try it out!

xxyzz:
I have started extract en edition, for over 2 hours now. There are too many outputs and I can't find the process estimate log so I don't know if it's faster or slower. But base on yesterday's tests I guess it's about the same as before.

xxyzz:
I finally find an estimate progress log, it's almost done, 8 minutes remain.

kristian-clausal:
That's wonderful! Hopefully this fixes the issues with the scaling.

xxyzz:
Actually it's 20 minutes faster than master.

PR #999, opened 2025-01-20
Talla2

kristian-clausal:
Some minor things from a failed bug-hunt. @xxyzz could you take a look at the `not wxr.config.analyze_template` thing? It looks like an oversight after something was changed, but I don't want to break it.

### DIFF
--- a/src/wiktextract/extractor/en/page.py
+++ b/src/wiktextract/extractor/en/page.py
@@ -2455,6 +2455,9 @@ def parse_language(
                 text, expand_all=True, template_fn=inflection_template_fn
             )
 
+            if not text.strip():
+                continue
+
             # Parse inflection tables from the section.  The data is stored
             # under "forms".
             if wxr.config.capture_inflections:
--- a/src/wiktextract/wiktwords.py
+++ b/src/wiktextract/wiktwords.py
@@ -464,7 +464,7 @@ def main():
                 wxr.wtp,
                 [Path(p) for p in args.override],
                 skip_extract_dump,
-                not wxr.config.analyze_templates,
+                None,
             )
 
         if args.page and not args.skip_extraction:


xxyzz (approved):
This is rarely used old code I forgot to update in #847, it's for overwriting pages when using the `--db-path` command option without the dump file path.  Non-en editions don't use "--override" option, they set pre-expand templates in `data/overrides` folder in JSON file.

Set to `None` is fine, unless you overwrite a template page and want to set more templates as pre-expand. But I think it's a mistake in en code because it expands too many templates accidentally and it's easier to ignore some templates like "trans-top".

And a small suggestion, I think `if len(text.strip()) == 0` is a bit more straightforward.

xxyzz:
Maybe we could remove the "--override" command option? Override page for debugging could be done with a short Python code to modify the .db file directly. Here is the code I use:

```python
import sqlite3
import sys

# usage: python this_file.py db_path page_path ns_id
# the last namespace id arg is optional
con = sqlite3.connect(sys.argv[1])
with open(sys.argv[2]) as f:
    title = sys.argv[2].replace("_", "/")
    if len(sys.argv) >= 4:
        ns_id = int(sys.argv[3])
    elif title.startswith("Module:"):
        ns_id = 828
    elif title.startswith("Template:"):
        ns_id = 10
    else:
        ns_id = 0
    print(f"{title=}")
    con.execute("INSERT INTO pages (title, namespace_id, body, need_pre_expand, model) values(?, ?, ?, 0, 'Scribunto') ON CONFLICT(title, namespace_id) DO UPDATE SET body=excluded.body, namespace_id=excluded.namespace_id", (title, ns_id, f.read()))

con.commit()
con.close()
```

kristian-clausal:
Let's, just leave it as is and remove it if it's needed.

Issue #973, opened 2025-01-06
Missing inflection templates for aii


ledzeppelin:
The 2025-01-03 kaikki dump from the enwiktionary dump dated 2025-01-01 for https://kaikki.org/dictionary/Assyrian%20Neo-Aramaic/index.html is missing `inflection_templates` for prepositions.

The XML file https://dumps.wikimedia.org/enwiktionary/20250101/enwiktionary-20250101-pages-articles.xml.bz2 contains the usages of inflection templates like Template:aii-infl-prep/1c but the kaikki dump does not.

The previous kaikki dump contained these templates. It would be really appreciated if you're able to look into this.

xxyzz:
I think it could be [Template:aii-infl-prep/1c](https://en.wiktionary.org/wiki/Template:aii-infl-prep/1c) table can't be extracted properly, pages use this template: https://en.wiktionary.org/wiki/Special:WhatLinksHere/Template:aii-infl-prep/1c

kristian-clausal:
I found the issue, the table had the header "Base Form" which wasn't in inflectiondata.py yet... Adding that fixed the issue, but I found a mysterious debug print that appeared when running the page so I'll hunt that down first.

ledzeppelin:
Hi @kristian-clausal, I just took a look at https://kaikki.org/dictionary/Assyrian%20Neo-Aramaic/index.html (off a wiktextract commit following the fix) - it looks like the issue is still occurring, would you be able to give it a second look?

kristian-clausal:
Please give us examples of where the error is occurring.

kristian-clausal:
Nevermind, I messed up... I seem to have had some sort of brainfart and added "Base Form" instead of "base form". But, the output on my computer is correct. I need to take a look.

kristian-clausal:
I rebuilt my database file from the latest dump and now I can recreate the error; yeah, that table is missing. There's been *some* changes to the table recently, so that may have broken it. The changes I made earlier were irrelevant, but I didn't check whether it worked *before* I did the changed (tut tut, me) so I thought it had fixed the issue.

kristian-clausal:
[...]

PR #1000, opened 2025-01-20
Revert "Merge pull request #986 from tatuylonen/split-templates"

ledzeppelin:
This reverts commit e6a833d3b47ae13db31d081e4aa99656302c3c43, reversing changes made to debb661f016f70096f4e9abeb7251332e686984a.


ledzeppelin:
@kristian-clausal - i think i figured out what the issue was, it has to do w/ the large values for `colspan` and `rowspan` which is an issue that can be fixed on the wiktionary side

i will need until the next xml dump feb 1st and the subsequent kaikki dump to confirm but let's revert since i dont think the entire parser should be modified bc of how 1 language implemented a couple of templates

ledzeppelin:
forgot to add another detail here... the reason this issue is only occurring for the preposition templates but not the noun templates - ie https://github.com/tatuylonen/wiktextract/issues/973#issuecomment-2587321471 is bc even though the noun templates set `rowspan="999"` it isn't a problem since the parent `<tr>` only contains `<th>` as opposed to a mix of `<th>` and `<td>`

kind of confusing but merging this pr and fixing on the wiktionary side of things is the way to go imo

kristian-clausal:
Do not start changing colspand and rowspan on Wiktionary side, it will only annoy the editors who explicitly have in the documentation that you're supposed to use "colspan=999" for this.

kristian-clausal:
This reversion is probably not necessary, because the changes made in the revert are still valid changes. The parsing of tables/template starts and ends specifically fixes a bug that's not specific to this issue. There should be no issue with keeping things like this. Please explain the issue with rowspan and colspan in more detail.

ledzeppelin:
> Do not start changing colspand and rowspan on Wiktionary side, it will only annoy the editors who explicitly have in the documentation that you're supposed to use "colspan=999" for this.

yah you're right, it does specifically mention that in the docs

let me close this PR

PR #1032, opened 2025-02-12
[en] remove "aii-infl-prep/1c" from "inflection_templates"

ledzeppelin:
GH issue https://github.com/tatuylonen/wiktextract/issues/973
Related PR https://github.com/tatuylonen/wiktextract/pull/1030

There's a lot aii templates which are afflicted by this issue and imo it's better if it works for all or none - otherwise it creates some confusion


xxyzz:
OK... but you still need "at-prep-auto" template?

xxyzz:
I've reverted pr 1030, I think it doesn't fix the problem but only a temporary fix.

ledzeppelin:
Sounds good, thanks for reverting

No I don't need ar-prep-auto

PR #1042, opened 2025-02-19
[en] Fix exception, dict is not pydantic object

kristian-clausal:
Also commented out two error message that were spamming a bit.

### DIFF
--- a/src/wiktextract/extractor/en/linkages.py
+++ b/src/wiktextract/extractor/en/linkages.py
@@ -1089,5 +1089,5 @@ def extract_l_template(
         elif span_lang.endswith("-Latn") and len(forms) > 0:
             roman = clean_node(wxr, None, span_tag)
             if roman != "":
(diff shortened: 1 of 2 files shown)


src/wiktextract/extractor/en/linkages.py
```diff
@@ -1089,5 +1089,5 @@ def extract_l_template(
         elif span_lang.endswith("-Latn") and len(forms) > 0:
             roman = clean_node(wxr, None, span_tag)
             if roman != "":
-                forms[-1].roman = roman
+                forms[-1]["roman"] = roman
```
xxyzz, line 1092: Sorry about the error...

kristian-clausal: Muscle memory is a powerful force, I know I would do this mistake (and I *will* do this mistake) all the time if I had to switch modes between Pydantic and [en]'s dicts.

PR #1044, opened 2025-02-19
[sv] extract gloss lists

xxyzz:


kristian-clausal:
Did you talk with Tatu about handling Swedish? It was one I was thinking of handling in the future because it's one of my mothertongues.

PR #1094, opened 2025-04-01
[el] Support parsing Linkage in Sense

daxida:
Partially addresses #1087 

I would like some feedback before continuing, specially since I feel like I am reinventing the wheel, and that some of my helper functions are probably already somewhere else.

Pasting the `mktest_bl_linkage` testing helper function docstring, these should be covered now. By linkage, read only (for now) βλ-linkage.
- [x] linkage inside synonym
- [x] linkage without gloss
- [x] linkage without gloss, but one quote
- [x] linkage without gloss, with βλ-prefixed linkage
- [x] linkage basic
- [x] linkage multiple
- [x] linkage useless (delete)

### DIFF
--- a/src/wiktextract/extractor/el/page.py
+++ b/src/wiktextract/extractor/el/page.py
@@ -41,7 +41,11 @@ from .section_titles import (
 def parse_page(
     wxr: WiktextractContext, page_title: str, page_text: str
 ) -> list[dict[str, WordEntry]]:
-    """Parse Greek Wiktionary (el.wiktionary.org) page."""
+    """Parse Greek Wiktionary (el.wiktionary.org) page.
+
+    References:
+    * https://el.wiktionary.org/wiki/Βικιλεξικό:Δομή_λημμάτων
+    """
 
     if wxr.config.verbose:
         logger.info(f"Parsing page: {page_title}")
@@ -123,7 +127,6 @@ def parse_page(
         )
 
         prev_data: WordEntry | None = None
-
 
         if len(sublevels) == 0 and ok:
             # Someone messed up by putting a Level 1 directly after a language
--- a/src/wiktextract/extractor/el/pos.py
+++ b/src/wiktextract/extractor/el/pos.py
@@ -9,7 +9,7 @@ from wikitextprocessor import (
     TemplateNode,
     WikiNode,
 )
-from wikitextprocessor.parser import LEVEL_KIND_FLAGS
+from wikitextprocessor.parser import LEVEL_KIND_FLAGS, WikiNodeChildrenList
 
 from wiktextract import WiktextractContext
 from wiktextract.page import clean_node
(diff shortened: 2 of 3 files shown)


kristian-clausal:
Thank you for your contribution; I'll continue working on this issue tomorrow, hopefully I can do something locally in the code that works with examples and glosses. Or, if every instance of `βλ` should be put into related, that shouldn't be too difficult.

daxida:
Answering to both your reply here and in the discussion. Do you then prefer me making an issue from the discussion, or should we just simply keep the conversation in this PR?

>  Or, if every instance of βλ should be put into related, that shouldn't be too difficult.

I think that should be the case. I am not aware of any counter-examples. Sometimes it makes sense to be deleted altogether though.

As far as I understand, other similar templates should receive the same treatment.

---

If you will work on this, I will try to be as little disruptive as I can. I could, for instance, focus only on simple tests. I don't know what is your policy on that, I see few in number tests in other languages, relatively bigger in size compared to what I did. I wonder if you'd prefer that format.

Let me know how you'd want me to proceed.

kristian-clausal:
Your tests and much of the code here is good, and it would obviously work fine, but there's a simpler way to (idiomatically in wiktextract) to extract template data: when we call things like `clean_node`, we can give an function argument for `template_fn=`, `post_template_fn=` or `node_handler_fn=`, which all work by basically receiving either a template and its arguments or a node, and then does things with them.

In this case, we could make a `template_fn` that intercepts the templates we want to extract from. The catch is that for the `template_fn` to communicate outside its scope we have to use a closure, so it'll be a sub-function with a non-local reference to a list that we can populate.

I'll see if I can do this easily by pulling this PR (I am still, after all these years, a git newbie at heart). Much of the logic and the helper functions you added in `pos.py` are, as you feared, probably not needed.

As a run-down of the different handler-functions:

* `template_fn` is called just before expanding any template, and you can intercept specific templates here. Returns either `None` (expand the original template as usual into text and insert normally) or a string (insert this in place of the template); returning "" basically removes the template.
* `post_template_fn` is the same, but happens after the template has been expanded and so you get the expanded text as a parameter so that you don't need to expand the template by hand if you need that text. Sometimes used, mostly a convenience function, I think.
* `node_handler_fn` does the same, except for any WikiNode, so this happens earlier than calling `template_fn`. Returns either `None`, a WikiNode, a string or a list of WikiNodes and/or strings, which replaces the handled node and its children when parsing the tree (but doesn't modify the original tree).

daxida:
Oops I missclicked that "ready for review".

Also I read your reply when I was about to commit that test (which ended up not being as simple as I initially thought). I will definitely check those functions.

> I'll see if I can do this easily by pulling this PR (I am still, after all these years, a git newbie at heart). Much of the logic and the helper functions you added in pos.py are, as you feared, probably not needed.

I don't know if merging it, then iterating on that, would have been more convenient for you. I can set it (with no missclick this time) to ready for review. It will polute a bit the git history but maybe that is fine?

kristian-clausal:
Just a quick point: the tests we write don't have access to a database of articles, so templates do not expand correctly, unless they're being intercepted. For example `# {{ετ|μτφρ|00=-}} {{βλ|καραγκιόζης}}` the et-template here should just produce an error message. You can add manually pages to the test's local instance if a template, module or article is needed using `wrx.wtp.add_page()`.

PR #1099, opened 2025-04-03
[el] Continuation from #1094 in a separate PR because I don't want to mess stuff up in git; "bl"-template handling

kristian-clausal:
Continuing from #1094

### DIFF
--- a/src/wiktextract/extractor/el/pos.py
+++ b/src/wiktextract/extractor/el/pos.py
@@ -9,7 +9,7 @@ from wikitextprocessor import (
     TemplateNode,
     WikiNode,
 )
-from wikitextprocessor.parser import LEVEL_KIND_FLAGS, WikiNodeChildrenList
+from wikitextprocessor.parser import LEVEL_KIND_FLAGS
 
 from wiktextract import WiktextractContext
 from wiktextract.page import clean_node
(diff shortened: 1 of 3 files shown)


kristian-clausal:
@daxida if you have time, could you take a look at this and see if it makes sense to you? I might have overlooked something in the implementation. Some of the tests have been changed.

src/wiktextract/extractor/el/pos.py
```diff
@@ -498,40 +498,8 @@ def bold_node_fn(
...
-                    parent_sense.raw_tags.append(LABEL_TAGS[arg_value])
-                elif isinstance(raw_tags, list):
-                    parent_sense.raw_tags.extend(LABEL_TAGS[arg_value])
-            contents[idx] = ""
-
-
 def parse_gloss(
-    wxr: WiktextractContext, parent_sense: Sense, contents: WikiNodeChildrenList
+    wxr: WiktextractContext, parent_sense: Sense, contents: list[str | WikiNode]
 ) -> bool:
```
daxida, line 503: Should I prefer that custom type? I was using `WikiNodeChildrenList` since it was already defined somewhere else.

tests/test_el_glosses.py
```diff
@@ -156,6 +170,11 @@ def test_bl_linkage_multiple(self) -> None:
                     "related": [
                         {"word": "ισοβαρής γραμμή"},
```
daxida, line 171: I think by looking at the wiki, it makes more sense if the related words were all be under the same gloss.

image

daxida (commented):
I like the changes but I would address the `test_bl_linkage_multiple` issue. Having two separate glosses does not seem right by looking at the wiki.

daxida:
As a side note, I was exploring the closures, and greping for `nonlocal` just finds results in `en` and `el`. Is there something specific to the Greek wikipage so that it needs to resort to this?

kristian-clausal:
The other extractors are programmed by xxyzz, who does things differently.

tests/test_el_glosses.py
kristian-clausal: Yes, but in another situation it might make more sense to have them separate. We can't create kludges for everything that Greek wiktionary does weirdly; the simplest fix here is to keep these as separate entries, and combine them later if needed. The data is still there.

src/wiktextract/extractor/el/pos.py
kristian-clausal: I think I wrote WikiNodeChildrenList for the purposes of Wikitextprocessor just because there were *so* many instances of `list[Union[str, WikiNode]]` everywhere in `parser.py`. I was meant as a local documentative type alias. There's nothing wrong with using it, but out of context `list[str | WikiNode]` is more accessible. You would need to hunt for the definition in wikitextprocessor otherwise.

daxida: Sure, if they are not meant to represent the same type then it makes sense.

Now for the hunting, I think every IDE should support some hovering action to expand the type.

image

kristian-clausal: I don't use an IDE that does that.

tests/test_el_glosses.py
kristian-clausal: You're right. Combining senses with identical `glosses` fields makes sense. I was hesitant to do it because it seemed like it would get really messy, especially merging senses together, but I later realized I could just make a custom `merge()` method for the pydantic object so that the logic for merging wouldn't be split up everywhere. You may have noticed that the code we have is less heavy on the use of classes than you'd expect, and mainly that's because Tatu is an old-school C-programmer and Wiktextract is the only real programming project I've ever really worked on. See #1102

daxida: Sorry for the late reply. Well, if that fixes I'm all for it, thank you. I was genuinely curious since there are so many patterns in the Wiki. It was very possible that I overlooked something.

Issue #1071, opened 2025-03-12
Word alignment in examples and translations


urroxyz:
On Wiktionary, example sentences are aligned with their entries.

Consider the third sense of "example":
> Something that serves as a [pattern](https://en.wiktionary.org/wiki/pattern#English) of [behaviour](https://en.wiktionary.org/wiki/behaviour#English) to be [imitated](https://en.wiktionary.org/wiki/imitate#English) (a [good](https://en.wiktionary.org/wiki/good#English) example) or not to be imitated (a [bad](https://en.wiktionary.org/wiki/bad#English) example).

The example text is as follows:
> Nelson Mandela was an **example** for many to follow.

In it, "**example**" is highlighted in bold.

It is important to know which word is associated with the entry due to inflections and polysemy.

Will there be any support for this in the near future? The best way to do this, in my opinion, would be to include a `raw_text`, `raw_roman` and `raw_english` property for each example, or a `raw_examples` collection altogether, each value with the original HTML.

Right now, I have written by own Python script to add the `<b>` and `</b>` tags back with web scraping, but this is inefficient and will eventually lead to API abuse.

xxyzz:
Maybe we could add the expanded wikitext, some templates like "ux" in page "example" only expand to wikitext not HTML. This feature might be impossible to implement in en code because it processes plain text and the node structure information are discarded.

urroxyz:
Thank you for the quick response!

Any quick fix would be fine as long as it grants the user a little bit of freedom, I think. Raw text, regardless of format, would be infinitely appreciated.

xxyzz:
BTW Wiktionary has HTML dump file: https://dumps.wikimedia.org/other/enterprise_html/runs/20250301/enwiktionary-NS0-20250301-ENTERPRISE-HTML.json.tar.gz so you don't have to abuse there API.

urroxyz:
Thank you so much again!

I couldn't find this, for some reason. Is there a home page with a list of the HTML dumps?

xxyzz:
https://dumps.wikimedia.org/other/enterprise_html/
https://dumps.wikimedia.org/other/enterprise_html/runs/

xxyzz:
I have created a commit to add bold word offsets for "zh-x" template in Chinese Wiktionary: https://github.com/tatuylonen/wiktextract/commit/dc2c4c0bfc7b9920976b01c2d67844c238e592bb (not merged yet)

[...]

PR #1103, opened 2025-04-04
[en, fr, zh] add example bold word offsets data

xxyzz:
#1071

Some templates expand to evil wikitext like `[[焚#Chinese|<b>焚]][[膏油#Chinese|膏</b>油]]` in page [焚膏繼晷](https://en.wiktionary.org/wiki/%E7%84%9A%E8%86%8F%E7%B9%BC%E6%99%B7), and our code can't handle this kind of node structure. Better fix the template Lua code on Wiktionary.

### DIFF
--- a/src/wiktextract/extractor/en/example.py
+++ b/src/wiktextract/extractor/en/example.py
@@ -18,7 +18,7 @@ def extract_example_list_item(
 ) -> list[ExampleData]:
     examples = []
     for template_node in list_item.find_child(NodeKind.TEMPLATE):
-        if template_node.template_name in ["zh-x", "zh-q"]:
+        if template_node.template_name in ["zh-x", "zh-usex", "zh-q"]:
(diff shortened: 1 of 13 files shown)


xxyzz:
I have tested some en edition pages and seemly I didn't break the code and also fix some example data in pages using "th-x" templates.

src/wiktextract/extractor/en/example.py
```diff
@@ -54,6 +55,25 @@ def extract_example_list_item(
...
+        elif template_node.template_name in [
+            "ux",
+            "usex",
+            "uxi",
+            "ko-usex",
+            "koex",
+            "ko-x",
+            "th-usex",
+            "th-x",
+            "th-xi",
```
xxyzz, line 67: These templates will be handled in this file in order to add the bold word offsets data.

PR #1108, opened 2025-04-08
[ru] extract "мест ru *" and "числ ru *" table templates

### LINKED ISSUE #1092: [es] We have found issues with some words in Spanish.
[Documento.docx](https://github.com/user-attachments/files/19547798/Documento.docx)

xxyzz:
Fix #1092


josunemarrtin03:
Thank you very much, we will continue to report if we see problems.

PR #1109, opened 2025-04-08
[de, es, ru] add example bold and italic word offsets data

xxyzz:


tests/test_es_example.py
```diff
@@ -46,6 +46,7 @@ def test_ejemplo_template(self):
                 {
                     "text": "And He has on His robe and on His thigh a name "
                     "written: KING OF KINGS AND LORD OF LORDS",
+                    "bold_text_offsets": [(74, 78), (82, 86)],
```
xxyzz, line 49: Currently `calculate_bold_offsets()` uses `re.finditer` to find bold word offsets but this will return an extra offset like in this test. Convert each node to text should fix this problem but I haven't figure out how to handle white spaces in each node.

PR #1131, opened 2025-04-16
[el] form of templates stuff

kristian-clausal:


kristian-clausal:
Now that it turns out that the `template_parameters` stuff was just in my head, this should be ok now.

PR #1134, opened 2025-04-17
[zh] improve translation subpage and linkage section code

xxyzz:


xxyzz:
The last commit also removes duplicated linkage data, previous code add them twice from https://github.com/tatuylonen/wiktextract/blob/530ae94ffb8f6524162aa296a8f640f7fe983ea5/src/wiktextract/thesaurus.py#L237 now only English Wiktionary extractor uses the thesaurus.py file.

PR #1151, opened 2025-04-28
[en] strip newline from template name

ledzeppelin:
this is regards to #973

when templates are used with newlines, ex. https://en.wiktionary.org/w/index.php?title=%DC%9F%DC%98%DC%A1%DC%AC%DC%AA%DC%90&action=edit&section=5 the name is not correctly parsed out, this attempts to fix that so we end up with `"aii-infl-noun"` instead of `"aii-infl-noun\n"` as the template name

### DIFF
--- a/src/wiktextract/extractor/en/analyze_template.py
+++ b/src/wiktextract/extractor/en/analyze_template.py
@@ -144,7 +144,7 @@ def analyze_template(wtp: Wtp, page: Page) -> tuple[set[str], bool]:
                  (\|  |  \}\})            # | or }}""",
         unpaired_text,
     ):
-        called_template = m.group(3)
+        called_template = m.group(3).strip()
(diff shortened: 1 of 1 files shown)


xxyzz:
The code at here is processing template pages not normal definition pages, the commit only remove white spaces of template names used in templates, probably not what you intended.

xxyzz:
The "aii-infl-noun" template is marked as need preexpanded in this code because it has partial wikitext table, that means it'll be expanded at here: https://github.com/tatuylonen/wiktextract/blob/b5667c2704304787a646e90af64690501ffbb8ec/src/wiktextract/extractor/en/page.py#L4425-L4430

and the `tree` node will contain the expanded table node not the original "aii-infl-noun" template node.

ledzeppelin:
myb thanks for clarifying, closing this

Issue #1180, opened 2025-05-10
Translations for English preposition  "in" are listed twice


ChristianSi:
I noticed a bug regarding the translations of the English word "in", which are collected on a separate [translation page](https://en.wiktionary.org/wiki/in/translations#Preposition). In Kaikki, all translations for the "Preposition" senses (e.g. "contained by" and "surrounded by") are listed twice, once with "prep" as POS (correct) and then again with "adj" as POS (wrong). I didn't check whether the Adverb and Noun senses are likewise affected, but I suspect so.

I noticed this in the JSONL file, but it's also noticeable on the [Kaikki page](https://kaikki.org/dictionary/English/meaning/i/in/in.html), where the translations for e.g. the "contained by" sense appear first in the Adjective entry with the note "The following are not (yet) sense-disambiguated" (wrongly) and then again in the Preposition entry under "Used to indicate location, inclusion, or position within spatial, temporal or abstract limits" (correctly).

xxyzz:
The last "Adjective" section should have a "Translations" section node, fixed on Wiktionary.

ChristianSi:
@xxyzz I see, thanks a lot!

kristian-clausal:
I'm currently taking a look at the en-edition translation page code to see if I we can make this a bit better, but I just spent like an hour being bewildered by a badly formatted error message that made me think there was some huge bug in the background, but turned out to be just that there was a format-editing function in-between that added a ton of json context to the error message and that the error message (and the code generating the error message) made me think there should have been a field when there wasn't...

xxyzz:
I always comment out this line locally https://github.com/tatuylonen/wiktextract/blob/0942e7da46d8b80b5613c0e208cffe3f485502f3/src/wiktextract/wiktionary.py#L186 to remove the full JSON data from check JSON error message, otherwise the error message is too damn long. Should we remove this line?

kristian-clausal:
Yeah, that honestly sounds fine.

kristian-clausal:
I've merged a PR (that I ran tests on locally, I am noting here if they break right now on github) that should add a bit of failsafe to these, with a list of possible "try these" paths for the translation subpage delving to go through if it doesn't find the correct thing at first.

PR #1189, opened 2025-05-13
Improvements on translation pages wip + error message changes in wiktionary.py

kristian-clausal:


src/wiktextract/wiktionary.py
```diff
@@ -289,16 +289,30 @@ def check_str_fields(
...
-                    "{!r} should be a{} string (it is a "
-                    "mandatory field): {}".format(
+                    "{!r} is missing and should be a{} string: {}".format(
+                        field,
+                        "" if empty_ok else " non-empty",
+                        json.dumps(item, sort_keys=True, ensure_ascii=False),
+                    ),
+                )
+            continue
+        if field is None:
```
xxyzz, line 307: `fields` list never contain `None`, IMO changes after this line could be removed.

kristian-clausal: oh yeah, of course, that makes sense. I didn't think about it, just did it.

kristian-clausal:
If I add the text

Fixes #1180 

Will github close that issue? Let's find out!

PR #1192, opened 2025-05-14
[el] Support `form_of` for `κλ` template

daxida:
Support extracting the `form_of` for the [κλ](https://el.wiktionary.org/wiki/Πρότυπο:κλ) generic (noun, verbs etc.) template.

Fixed a couple typos too, and updated the documentation.

The git diff is a bit loud because my formatter didn't like the wxr warnings in two lines.

Also I am not sure what the `sortid` of the warnings accomplishes, I just copied the one a couple lines below. Let me know if I should change it.

### DIFF
--- a/src/wiktextract/extractor/el/pos.py
+++ b/src/wiktextract/extractor/el/pos.py
@@ -515,8 +515,15 @@ def extract_form_of_templates(
 ) -> None:
     """Parse form_of for nouns, adjectives and verbs.
 
+    Supports:
+    * κλ             | generic                | form_of
+    * πτώση/πτώσεις  | nouns, adjectives etc. | form_of and tags
+    * ρημ τύπος      | verbs                  | form_of
+
     * References:
+    https://el.wiktionary.org/wiki/Πρότυπο:κλ
     https://el.wiktionary.org/wiki/Κατηγορία:Πρότυπα_για_κλιτικούς_τύπους
+    https://el.wiktionary.org/wiki/Πρότυπο:ρημ_τύπος
     """
 
     # For the moment...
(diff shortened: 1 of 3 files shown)


src/wiktextract/extractor/el/pos.py
```diff
@@ -531,23 +538,38 @@ def extract_form_of_templates(
...
     t_name = t_node.template_name
+
+    # Generic
+    if t_name == "κλ":
+        t_args = t_node.template_parameters
+        if 2 not in t_args:
+            wxr.wtp.warning(
+                "Form-of template does not have lemma data: "
+                f"{t_name}, {t_args=}",
+                sortid="pos/535/20250416",
```
kristian-clausal, line 550: sortid is not critical, but it's nice to have: we use it on the kaikki.org website to sort all the error messages so that messages generated in one place in the code are next to each other, basically it's just an id for the error.

kristian-clausal (approved):

src/wiktextract/extractor/el/pos.py
daxida: If the sortid could be shared among warnings of that function (and of the πτώση extraction), it may be interesting to export the warning to a `form_of_warning` function to avoid some boilerplate/noise.

kristian-clausal: It's meant for sorting, mostly, and for finding where the error/warning/debug message was called from, including location inside a function, if it's unclear. Giving them unique strings should be enough.

daxida: I see. Should then I just update the date for the sortid used in the "κλ" template to distinguish it from the `ρημ τύπος`. If I understand correctly it's line number then date, so `sortid="pos/550/20250514"`?

kristian-clausal: Sorry, I was quick to merge, and now it's complicated; you don't have to do it, but yeah, the format doesn't matter, just that it's unique (for convenience). I use "filename/line" because it's easy, and then the date when there's a chance of collision, but now that I've merged it you don't need to bother, I'll do it in some other commit if I remember.

Issue #1185, opened 2025-05-12
[el] Skipping `:` nodes in when listing glosses


daxida:
Many participles (μετοχή) have their `see` (`βλ`) redirection skipped because they follow a simple pattern with only one element listed after a colon (cf. [εκφρασμένος](https://el.wiktionary.org/wiki/%CE%B5%CE%BA%CF%86%CF%81%CE%B1%CF%83%CE%BC%CE%AD%CE%BD%CE%BF%CF%82), [αναμαλλιασμένος](https://el.wiktionary.org/wiki/%CE%B1%CE%BD%CE%B1%CE%BC%CE%B1%CE%BB%CE%BB%CE%B9%CE%B1%CF%83%CE%BC%CE%AD%CE%BD%CE%BF%CF%82) and many others from [this list](https://el.wiktionary.org/wiki/%CE%9A%CE%B1%CF%84%CE%B7%CE%B3%CE%BF%CF%81%CE%AF%CE%B1:%CE%9C%CE%B5%CF%84%CE%BF%CF%87%CE%AD%CF%82_%CF%80%CE%BF%CF%85_%CE%BA%CE%BB%CE%AF%CE%BD%CE%BF%CE%BD%CF%84%CE%B1%CE%B9_%CF%8C%CF%80%CF%89%CF%82_%CF%84%CE%BF_%27%CE%B1%CE%B3%CE%B1%CF%80%CE%B7%CE%BC%CE%AD%CE%BD%CE%BF%CF%82%27_(%CE%BD%CE%AD%CE%B1_%CE%B5%CE%BB%CE%BB%CE%B7%CE%BD%CE%B9%CE%BA%CE%AC)).

```
==={{μετοχή|el}}===
'''{{PAGENAME}}, -η, -ο'''
: {{βλ|εκφράζω}}
{{clear}}
```

This is trivially fixed by removing this line:

https://github.com/tatuylonen/wiktextract/blob/2cb1898bf57d30ce446a409bd7e02d8b0294b414/src/wiktextract/extractor/el/pos.py#L120

this produces the intended result on a test that I wrote, but unfortunately it makes this other test crash.

https://github.com/tatuylonen/wiktextract/blob/2cb1898bf57d30ce446a409bd7e02d8b0294b414/tests/test_el_head.py#L70

Because the `forms` entry is not longer there, since whe increased the `glosses_index` when parsing `glosses_lists`.

[...]

PR #1194, opened 2025-05-14
[el] Test ":" list with a single item

daxida:
Simple test following the discussion at #1185


kristian-clausal:
Thank you for contributing! A++ would merge again

PR #1267, opened 2025-07-08
[en] don't extract "sense" and "color panel" templates as example

xxyzz:
example page "outer space", "twit", "emancipate"


kristian-clausal (approved):

kristian-clausal:
This seems correct, thanks for checking this out!

PR #1279, opened 2025-07-22
[doc] add create new extarctor guide

### LINKED ISSUE #205: Guide to adding new Wiktionaries
Hi everyone,

I thought it would be really cool to have a guide to add new other-language Wiktionaries. I was trying to do some [work on the Russian Wiktionary ](https://github.com/Vuizur/wiktextract/tree/russian-wiktionary) and wrote my current understanding down.

# Guide to adding new languages

1. Create the necessary language data in wiktextract/data. 
    * First you need to create `languages.json`, mapping language codes to language names. (For the English Wiktionary, this would map "en" to "English"). The generation of these files is handled in `get_languages.py`. Depending on the Wiktionary, the best way to do this varies, you either have to expand some templates or parse source code. (Simply look at the examples.)
    * Manually create:
        * `pos_subtitles`: The translated names of the parts of speech. (For Russian Wiktionary this might be problematic, because the POS are at the beginning of a string containing all sorts of grammar information, so one would have to use a regex or so)
        * `linkage_subtitles.json`: Contains the translated names of synonym/antonym/... sections
        * `other_subtitles.json`:  Has the translated names of the inflection/etymology sections (lower case)
        * `zh_pron_tags.json`: Not sure what it does exactly, but the file has to exist and contain an empty dictionary `{}` at least
        * `form_of_templates.json`: 

2. Run the program using 

```bash
wiktwords --all --all-languages --out data.json --dump-file-language-code <yourlangcode> enwiktionary-20201201-pages-articles.xml.bz2 
```

For Russian: `wiktwords --all --all-languages --out data.json --dump-file-language-code <yourlangcode> ruwiktionary-20230101-pages-articles-multistream.xml` 

I am currently still a bit confused about the "compounds" key in `other_subtitles.json`. What section exactly does it refer to? I cannot seem to find it in the Russian Wiktionary.

@xxyzz

xxyzz:
The rendered markdown file is more easier to read, it can be viewed by clicking the view file button:

image

it also has table of contents: 
image

close #205

### DIFF
--- a/docs/new_extractor_guide.md
+++ b/docs/new_extractor_guide.md
@@ -1,0 +1,1559 @@
+# Create new Wiktionary extractor guide
+
+## Learn wikitext
+
+Read document [Help:Wikitext](https://en.wikipedia.org/wiki/Help:Wikitext) to learn the basic syntax of wikitext.
+
+More MediaWiki documents:
+- [Help:A quick guide to templates](https://en.wikipedia.org/wiki/Help:A_quick_guide_to_templates)
+- [Help:Template](https://en.wikipedia.org/wiki/Help:Template)
+- [Help:Transclusion](https://en.wikipedia.org/wiki/Help:Transclusion)
+- [Help:Extension:ParserFunctions](https://www.mediawiki.org/wiki/Help:Extension:ParserFunctions)
+- [Help:Magic words](https://www.mediawiki.org/wiki/Help:Magic_words)
+- [Extension:Scribunto/Lua reference manual](https://www.mediawiki.org/wiki/Extension:Scribunto/Lua_reference_manual)
+
+## Read Wiktionary entry layout document
+
+Every Wiktionary has their unique page layout, they usually have a document page of the layout and other rules. For example, here is English Wiktionary's document: [Wiktionary:Entry layout](https://en.wiktionary.org/wiki/Wiktionary:Entry_layout). After reading the layout document, try edit some pages or create a new page to test whether you have grasped Wiktionary editing. Writing Wiktionary extractor is like a reversed editing process.
+
+## Learn wikitextprocessor parser API
+
+[wikitextprocessor](https://github.com/tatuylonen/wikitextprocessor) is a wikitext parser that takes in wikitext and generates a parse-tree.
+
+Here is a simplified wikitext of English Wiktionary page [dictionary](https://en.wiktionary.org/wiki/dictionary) and we'll use wikitextprocessor library to parse it:
+
+```wikitext
+==English==
+===Noun===
+# {{lb|en|computing}} An [[associative array]]
+```
+
+```python
+from wikitextprocessor import Wtp
+from wiktextract import WiktionaryConfig, WiktextractContext
+
+wtp = Wtp(db_path="db_path", lang_code="en")
+conf = WiktionaryConfig(
+    dump_file_lang_code="en", capture_language_codes=None
+)
+wxr = WiktextractContext(wtp, conf)
+wxr.wtp.start_page("dictionary")
+root = wtp.parse(
+    """==English==
+===Noun===
+# {{lb|en|computing}} An [[associative array]]"""
+)
+```
+
+The level 3 Part of Speech (POS) node(`===Noun===`) is the child of level 2 language node(`==English==`):
+
+```python
+from wikitextprocessor.parser import print_tree
+
+print_tree(root)
+# ROOT [['']]
+#  LEVEL2 [['English']]
+#    '\n'
+#    LEVEL3 [['Noun']]
+#      '\n'
+#      LIST #
+#        LIST_ITEM #
+#          ' '
+#          TEMPLATE [['lb'], ['en'], ['computing']]
+#          ' An '
+#          LINK [['associative array']]
+print(root.children)
+# [<LEVEL2(['English']){} '\n', <LEVEL3(['Noun']){} '\n', <LIST(#){} <LIST_ITEM(#){} ' ', <TEMPLATE(['lb'], ['en'], ['computing']){} >, ' An ', <LINK(['associative array']){} >>>>>]
+```
+
+Child node can be located with these APIs: `WikiNode.find_child()`, `WikiNode.find_child_recursively()`, `WikiNode.invert_find_child()`, `WikiNode.find_html()`, `WikiNode.find_html_recursively()`. Parser API source code: [src/wikitextprocessor/parser.py](https://github.com/tatuylonen/wikitextprocessor/blob/main/src/wikitextprocessor/parser.py)
+
+```python
+from wikitextprocessor import NodeKind
+
+for t_node in root.find_child_recursively(NodeKind.TEMPLATE):
+    pass
+```
+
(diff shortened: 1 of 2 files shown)


docs/new_extractor_guide.md
```diff
@@ -0,0 +1,1531 @@
...
+    base_data: WordEntry,
+    level_node: LevelNode,
+    pos_title: str,
+) -> None:
+    page_data.append(base_data.model_copy(deep=True))
+    page_data[-1].pos_title = pos_title
+    pos_data = POS_DATA[pos_title]
+    page_data[-1].pos = pos_data["pos"]
+    page_data[-1].tags.extend(pos_data.get("tags", []))
+
```
kristian-clausal, line 332: `# Extract glosses: lists starting with "#" with senses for the words.

docs/new_extractor_guide.md
```diff
@@ -0,0 +1,1531 @@
...
+    page_data: list[WordEntry],
+    base_data: WordEntry,
+    level_node: LevelNode,
+    pos_title: str,
+) -> None:
+    page_data.append(base_data.model_copy(deep=True))
+    page_data[-1].pos_title = pos_title
+    pos_data = POS_DATA[pos_title]
+    page_data[-1].pos = pos_data["pos"]
+    page_data[-1].tags.extend(pos_data.get("tags", []))
```
kristian-clausal, line 331: Maybe add a comment about parsing headers here, something like "if the parsed wiktionary edition has a section before glosses that contains information about the word, called a 'head' in English, it would be handled here".

docs/new_extractor_guide.md
```diff
@@ -0,0 +1,1531 @@
...
+    for span_tag in expanded_node.find_html_recursively(
+        "span", attr_name="class", attr_value="label-content"
+    ):
+        sense.raw_tags.append(clean_node(wxr, None, span_tag))
+    # save categories
+    clean_node(wxr, sense, expanded_node)
+```
+
+There are three ways to extract a template:
+- expand template, find data in expanded nodes: this method is used when data in only available when template is expanded or it's easier to extract data from node structure, for example, data may already separated in different nodes
```
kristian-clausal, line 376: Maybe swap these two lines so that "don't expand template" is first. That's the order you'd do it if you use both methods, first check parameters and then expand? This is a very minor nitpick.

docs/new_extractor_guide.md
```diff
@@ -0,0 +1,1531 @@
...
+- don't expand template, find data in template parameters: this is used when a template only display the parameter and does nothing more than that. If a template expand nodes contain category links, it's still need to be expanded to get the category data
+- convert template node to text using `clean_node`: this is for expanded template text can be used directly or need minimal change
+
+The first and second methods are sometimes used together when some data are easier to obtain from expanded nodes and some can be obtained directly from parameter without much change.
+
+## Create "tags.py" file
+
```
kristian-clausal, line 383: "What are tags" section could be added here. "tags" are the English-language linguistics terms that should be common and shared between all wiktextract editions, etc. etc. and raw_tags are things we don't know about yet.

docs/new_extractor_guide.md
```diff
@@ -0,0 +1,1531 @@
...
+    data.raw_tags = raw_tags
+```
+
+- English edition [sense tags](https://kaikki.org/dictionary/errors/mapping/index/senses/_list_/tags.html)
+- English edition [form tags](https://kaikki.org/dictionary/errors/mapping/index/forms/_list_/tags.html)
+- English edition [topics](https://github.com/tatuylonen/wiktextract/blob/master/src/wiktextract/topics.py)
+
+## Add test
+
+Create test file in "tests" folder. For example, [tests/test_it_gloss.py](https://github.com/tatuylonen/wiktextract/blob/master/tests/test_it_gloss.py). Tests should be included in the same git commit along with the extractor code, so it'll be more helpful for understanding the code when using git blame.
```
kristian-clausal, line 412: I will try to be better... orz

docs/new_extractor_guide.md
```diff
@@ -0,0 +1,1531 @@
...
+                            "categories": ["English transitive verbs"],
+                            "tags": ["transitive"]
+                        }
+                    ],
+                }
+            ],
+        )
+```
+
+`Wtp.add_page()` could be used to add templates required in test. If different expanded outputs are needed, we can use the [`#switch`](https://www.mediawiki.org/wiki/Help:Extension:ParserFunctions##switch) wikitext parser function to control the output.
```
kristian-clausal, line 494: This is neat.

kristian-clausal (commented):
Looking excellent.

docs/new_extractor_guide.md
```diff
@@ -0,0 +1,1531 @@
...
+        validate_default=True,
+    )
+
+class Sense(EnglishBaseModel):
+    glosses: list[str] = []
+    categories: list[str] = []
+    topics: list[str] = []
+    tags: list[str] = []
+    raw_tags: list[str] = []
+
```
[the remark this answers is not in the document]
xxyzz: Two comments added above the class declare lines.

docs/new_extractor_guide.md
```diff
@@ -0,0 +1,1531 @@
...
+    pos: str = Field(default="", description="Part of speech type")
+    pos_title: str = Field(
+        default="", description="Original POS title"
+    )
+    senses: list[Sense] = Field(default=[], description="Sense list")
+```
+
+## Create "section_titles.py" file
+
+This file contains POS and other section titles data. Example file from English Wiktionary extractor: [src/wiktextract/extractor/en/section_titles.py](https://github.com/tatuylonen/wiktextract/blob/master/src/wiktextract/extractor/en/section_titles.py)
```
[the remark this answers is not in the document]
xxyzz: The first full Part of Speech name is mentioned in "wikitextprocessor api" section since it's moved now and the rest texts use "POS" abbreviation. I think readers could understand wikitext section after reading the documents listed in the first two sections.

docs/new_extractor_guide.md
```diff
@@ -0,0 +1,1531 @@
...
+```python
+root = wxr.wtp.parse('<span class="span_class">text</span>')
+for span_tag in root.find_html("span"):
+    print(span_tag.tag)
+    # "span"
+    print(span_tag.attrs)
+    # {"class": "span_class"}
+```
+
+List type can be distinguished by checking the `WikiNode.sarg` attribute:
```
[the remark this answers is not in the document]
xxyzz: I think we already renamed or changed these fields once? Rename could be tiresome...

docs/new_extractor_guide.md
xxyzz: Comment added

docs/new_extractor_guide.md
xxyzz: Explanation added.

kristian-clausal (approved):
This looks good. Most quibbles have been taken care of, no major typoes that could cause misunderstandings that I could see.

xxyzz:
Thanks for the review! I'll rebase and squash the commits than merge this pr.

Issue #1278, opened 2025-07-18
[en] "defdate" template data doesn't get processed.


praxiq:
Many definitions in Wiktionary have a [`defdate`](https://en.wiktionary.org/wiki/Template:defdate) that indicates when that meaning came into usage. It's defined with a template, like `{{defdate|ca. 1480}}` or `{{defdate|from 1570s}}`. The extract seems to totally discard this info.

Some example lines from the definition of ["dictionary"](https://en.wiktionary.org/wiki/dictionary) (see the end of each line):

```wikitext
# {{senseid|en|Q23622}} A [[reference work]] listing [[word]]s or [[name]]s from one or more [[language]]s, usually ordered [[alphabetically]], explaining each word's [[meaning#Noun|meanings]] or {{l|en|sense|id=any particular meaning of a word|senses}}, oftentimes also containing information on its [[etymology#Noun|etymology]], [[pronunciation#Noun|pronunciation]], {{l|en|usage|id=language}}, [[Wiktionary:Semantic relations|semantic relations]], [[translation#Noun|translations]], as well as other relevant information. {{defdate|ca. 1480}}
# {{lb|en|by extension}} A [[reference work]] on a particular [[subject]] or [[activity]] in which the entries are arranged alphabetically; an [[alphabetical]] [[encyclopedia]]. {{defdate|from 1570s}}
```

xxyzz:
This template was removed from gloss text several years ago, and I agree with this change. We'll probably need a new field for this data, or maybe put it in "raw_tags".

`sense_template_fn()` at here: https://github.com/tatuylonen/wiktextract/blob/45c4a21f91e902e6390787e5e17aea7cfc4b9064/src/wiktextract/extractor/en/page.py#L1941-L1942

make it difficult to expand the template, the template node is inaccessible. This why I don't like `template_fn` argument of `clean_node()`.

IMO, iterate through the gloss list child nodes is more intuitive and flexible.

kristian-clausal:
A new field for a sense's earliest attestation isn't a bad idea. `defdate` is not a good name for it, because the sense hasn't necessarily been "defined" by anyone, it's just the earliest time the sense has been found in text or recordings.

Not a high priority at the moment. Tatu hasn't historically been interested in the etymological data that much, and I tried to convince him that it was important, but the code related to this template is obviously from before that.

[...]

PR #1301, opened 2025-08-08
[en] Add "attestations" field to senses (Template:defdate)

kristian-clausal:
Issue #1278

Done within the gloss/sense template handler function.

`Template:defdate` is used to show the earliest (or just early) attested dates in text where a word, or a word sense, has been found. It's main parameter (#1) is a string data like "early 19th century", and it can have up to three `refX=` and `refXn=` arguments. The ref (based on how it's in etymology templates) is probably a `Template:ref` style reference that is then output by the Wiktionary expander with a `<references/>` tag later in the article, while the `refn` arguments refer to these without repeating the `ref` part. I think, based on a quick look, that we don't do special reference handling with `ref` template data, but the same data can be found within the word data so users should be able to access it.

For example in the article `gratis` for Swedish, there's a `ref` in the etymology section which is later refered to by `refn` arguments in later `defdate` templates. Both pieces are available and the `refn` should be unambiguous.

### DIFF
--- a/src/wiktextract/extractor/en/page.py
+++ b/src/wiktextract/extractor/en/page.py
@@ -81,6 +81,7 @@ from .section_titles import (
 )
 from .translations import parse_translation_item_text
 from .type_utils import (
+    AttestationData,
     DescendantData,
     ExampleData,
     FormData,
@@ -1931,6 +1932,23 @@ def parse_language(
                         return info_exp
                     return ""
             if name in ("defdate",):
+                date = clean_node(wxr, None, ht.get(1, ()))
+                refs = []
+                named_refs = []
+                for k, v in ht.items():
+                    if isinstance(k, str) and "ref" in k:
+                        ref_v = clean_node(wxr, None, v)
+                        if "n" in k:
+                            named_refs.append(ref_v)
+                        else:
+                            refs.append(ref_v)
+                data_append(
+                    sense_base,
+                    "attestations",
+                    AttestationData(
+                        date=date, refs=refs, refns=named_refs
+                    ),
+                )
                 return ""
             if name == "senseid":
                 langid = clean_node(wxr, None, ht.get(1, ()))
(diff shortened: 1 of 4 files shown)


kristian-clausal:
Won't merge on Friday afternoon.

xxyzz:
A more common use case is not considered: `{{defdate|17th|19th c.}}` example in the template [document page](https://en.wiktionary.org/wiki/Template:defdate) expands to "[17th–19th c.]".

Also reference names are not guaranteed to be added: `{{defdate|1|2|ref=ref1|ref2=ref2|ref2n=description 2}}`, here the name is only added to the second reference. This example is made up by me, I don't know if it actually exists. But the above second argument case have real usages.

IMO expand the template is better than reimplement it in Python...

kristian-clausal:
I don't know how I missed that second case in the documentation... I'll move this to post_template_fn.

I was thinking that hopefully ref numbers would always be in sync, but I wasn't happy with the list approach because if the case you mentioned just so happens to happen... I think I'll make refs a list of Ref dicts with a text and refname field. In that case, other references should also be handled the same way, although the one in etymologies is handled as an etymology_template.

kristian-clausal:
I think I've handled now most of the `defdate` behavior correctly. I didn't end up using post_template_fn because 1) it made writing tests more difficult and 2) the expanded text would have required some extra parsing because of all the HTML and the included `ref` elements. No worth it, there is no underlying module being called and `defdate` is a purely template-oriented "fill in the blanks" sort of template, so there are no transformations taking place. TLDR, just cleaning the `[1]` and `[2]` arguments is enough.

I tried to figure out whether `ReferenceData` could be added to other fields; for example, when a `defdate` template is used in an etymology section it gets added to `etymology_templates` with its argument data, but... It feels like that is not necessary or easy to do. We don't collect reference data anywhere else (mmmmaybe we could), and starting to collect it would be a whole project.

PR #1302, opened 2025-08-11
[en] fix pronunciation data in page "大家"

xxyzz:
the two POS sections under two pronunciation sections should have different "sounds" data

the second test is for commit 24b7d98 which added the `inside_level_four` variable

I don't know whether the change make sense but it passes the tests...

### DIFF
--- a/src/wiktextract/extractor/en/page.py
+++ b/src/wiktextract/extractor/en/page.py
@@ -1140,11 +1140,11 @@ def parse_language(
         # print(f"{stack=}")
         if wxr.wtp.subsection is not None:
             return pos_data
+        if inside_level_four:
+            return level_four_data
         if stack[-1] == language:
             return base_data
-        if inside_level_four is False:
-            return etym_data
-        return level_four_data
+        return etym_data
 
     term_label_templates: list[TemplateData] = []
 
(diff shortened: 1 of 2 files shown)


src/wiktextract/extractor/en/page.py
```diff
@@ -1140,6 +1140,8 @@ def select_data() -> WordData:
         # print(f"{stack=}")
         if wxr.wtp.subsection is not None:
             return pos_data
+        if inside_level_four:
+            return level_four_data
         if stack[-1] == language:
             return base_data
         if inside_level_four is False:
```
kristian-clausal, line 1147: 1147-1149 could be replaced with `return etym_data`, right? Unless "inside_level_four" contains non-boolean truthy data... if inside_level_four four is True or False it never reaches 1149? I don't know why I had "inside_level_four is False" originally and "return level_four_data" as the default, seems weird. This looks fine to me, it seems that I had some kind of brain-fart here.

kristian-clausal (approved):

kristian-clausal (approved):

xxyzz:
Damn it... now pronunciation section before etymology section data are not added. Example page "中國", "日語", "tee".

PR #1312, opened 2025-08-15
Deprecated fields

kristian-clausal:
We are going to deprecate some differences between the newer extractors and the old English language extractor, mainly changing field names.

`english` in the meaning of "the english translation of this [...]" has been changed to "translation". The `english` field will remain in place for a while, and we'll make a deprecation thread about it (with these other changes), but please start using `translation` fields instead.

`code` has been deprecated in favor of `lang_code` similarly.

Other pull requests will have other changes.

### DIFF
--- a/src/wiktextract/extractor/en/page.py
+++ b/src/wiktextract/extractor/en/page.py
@@ -975,6 +975,8 @@ def parse_language(
     level_four_data: WordData = {}  # Chinese Pronunciation-sections in-between
     etym_data: WordData = {}  # For one etymology
     sense_datas: list[SenseData] = []
+    sense_ordinal = 0  # The recursive sense parsing messes up the ordering
+    # Never reset, do not use as data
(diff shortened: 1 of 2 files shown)


kristian-clausal:
Forgot to mention this, but I changed `bold_english_offsets` straight to `bold_translation_offsets`. @xxyzz if this seems to hasty, just say and I'll change to be deprecated instead. `bold_english_offsets` is pretty new, so people aren't probably using it that much yet..?

xxyzz (approved):

xxyzz:
I think it's fine, it's a new data added I guess a few months ago?

kristian-clausal:
I'm going to put the gloss-ordering stuff in this PR so I'll not merge yet. :+1:

kristian-clausal:
I can't believe that I haven't been able to figure out how to properly get the senses sorted. Next week! this shouldn't be this hard.

Issue #1314, opened 2025-08-17
[en] erroneous inflection data for "can" (verb) -> "canneth"


gold-silver-copper:
When processing data for my project, I came upon an improperly tagged inflection of the word "can", which is "canneth". In the data, "canneth" is not tagged as archaic. 

https://en.wiktionary.org/wiki/canneth
https://en.wiktionary.org/wiki/can#English

`
Entry {
    word: "can",
    pos: "verb",
    forms: Some(
        [
            Forms {
                form: "can",
                tags: [
                    "present",
                    "singular",
                    "third-person",
                ],
            },
            Forms {
                form: "able",
                tags: [
                    "participle",
                    "present",
                    "suppletive",
                ],
            },
            Forms {
                form: "could",
                tags: [
                    "past",
                ],
            },
            Forms {
                form: "couth",
                tags: [
                    "obsolete",
                    "participle",
                    "past",
                ],
            },
        ...
        ...
            Forms {
                form: "can",
                tags: [
                    "present",
                    "singular",
                    "third-person",
                ],
            },
            **Forms {
                form: "canneth",
                tags: [
                    "present",
                    "singular",
                    "third-person",
                ],
            },**
        ...
        ...
        ...
            Forms {
                form: "kin",
                tags: [
                    "alternative",
                    "pronunciation-spelling",
                ],
            },
        ],
    ),
    lang_code: "en",
}

`

xxyzz:
Form "canneth" is extracted from the conjugation table in page "can", not from the "canneth" page, so the "archaic" tag in  "canneth" page can't be added. But the table has a "Sparsely attested near 1500." note for the word "canneth" that could be added to "raw_tags" list.

gold-silver-copper:
> Form "canneth" is extracted from the conjugation table in page "can", not from the "canneth" page, so the "archaic" tag in "canneth" page can't be added. But the table has a "Sparsely attested near 1500." note for the word "canneth" that could be added to "raw_tags" list.

Ah okay I see it now, I missed the conjugation table in the entry for can. Thank you for the clarification, I will close this issue now.

kristian-clausal:
[...]

PR #1323, opened 2025-08-22
[en] Transform footnotes with `xlat_descs_map`

kristian-clausal:
Fixes issue #1314, 'canneth' form missing tags

Turns out that `xlat_descs_map` is used differently from what I thought: it is used to transform text straight into tag text, which is then usually parsed with decode_tags down the line. It works because the raw tag, even things like `with-dummy-subject` get decoded directly.

Should rename it. I don't even know what `xlat` means.

Anyhow, the fix here is that certain footnotes are just so messy we can't do much about them, so add them to this map and transform the text into tags with `map_with`.

This makes parsing footnotes slower, but that's such a small share of overall time spent that it shouldn't really matter.

I'm starting to understand why enterprise code names things verbosely.

### DIFF
--- a/src/wiktextract/extractor/en/form_descriptions.py
+++ b/src/wiktextract/extractor/en/form_descriptions.py
@@ -874,7 +874,7 @@ def distw(titleparts: Sequence[str], word: str) -> float:
 
 
 def map_with(
-    ht: Union[dict[str, Union[str, list[str]]], dict[str, str]],
+    ht: dict[str, str | list[str]] | dict[str, str],
     lst: Sequence[str],
 ) -> list[str]:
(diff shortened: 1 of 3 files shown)


xxyzz:
The text after "1500." is from a `<ref>` tag(`Sparsely attested near 1500.<ref>{{R:OED Online|id=4994066876|pos=v.<sup>1</sup>|date=December 2024}}</ref>`), doesn't `clean_node()` remove `<ref>` content? And if the argument of the template changes the expanded text also changes...

kristian-clausal:
That text gets all the way down to `add_def`, with the text generated by the references and all. You're right that the it shouldn't be there, I'll take a look at it.

gold-silver-copper:
You guys are awesome

PR #1391, opened 2025-10-06
[el] Support `form_of` for `γρ` template

daxida:
Add `form_of` for this template: [γρ](https://el.wiktionary.org/wiki/Module:%CE%AC%CE%BB%CE%BB%CE%B7%CE%BC%CE%BF%CF%81%CF%86%CE%AE)

The code would have been simpler if I didn't observe some users doing questionable things with the template. In their defense, the template documentation could have been clearer. It mixes alternatives writings with forms of, their example is completely wrong, it works when it should not, etc.

Nevertheless, in my experience this makes sense, and follows the same logic as the other `form_of` templates.

There is some noise due to my formatter. I hope that's not a problem.

### DIFF
--- a/src/wiktextract/extractor/el/pos.py
+++ b/src/wiktextract/extractor/el/pos.py
@@ -514,7 +514,7 @@ def bold_node_fn(
 
 def extract_form_of_templates(
     wxr: WiktextractContext,
-    parent_sense: Sense | WordEntry,
+    parent_sense: Sense,
     t_node: TemplateNode,
     siblings: list[str | WikiNode],
     siblings_index: int,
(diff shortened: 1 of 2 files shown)


src/wiktextract/extractor/el/pos.py
```diff
@@ -514,7 +514,7 @@ def bold_node_fn(
 
 def extract_form_of_templates(
     wxr: WiktextractContext,
-    parent_sense: Sense | WordEntry,
```
kristian-clausal, line 517: Do not remove type annotations. This was added by me recently.

src/wiktextract/extractor/el/pos.py
```diff
@@ -548,6 +550,17 @@ def extract_form_of_templates(
     if t_name == "κλ":
         return basic_extract(extract_argument=2)
 
+    # Notes:
+    # * All occurrences in wiktionary have at least one argument
+    # * Only handle cases where the second argument refers to a form:
+    #   μορφ / μορφή / λόγια μορφή του, etc.
+    #   and ignore those mistakenly used as synonym templates
+    if t_name == "γρ":
+        second_arg = t_node.template_parameters[2]
```
kristian-clausal, line 559: This breaks if there's one or zero arguments; use `.get()` instead and check it.

kristian-clausal (changes requested):

src/wiktextract/extractor/el/pos.py
daxida: Fixed, thanks

daxida:
Types seem fine now. I ran:

```
mypy src/wiktextract/extractor/el/**.py
Success: no issues found in 17 source files
```

kristian-clausal:
Thanks for your contribution! :+1:

Issue #1394, opened 2025-10-07
[el] Greek tags


daxida:
I'm sorry if I keep opening issues for a language that is most likely not on top of your priorities.

I remember @kristian-clausal saying he worked on this. I wonder if I could help, mainly by translating some tags.

I am interested in translating a small subset: grammar terms that appear in verb tables. Of the likes of `raw_tag: υποτακτική > tag: subjunctive` etc.

I am unsure where to start. `src/wiktextract/extractor/el/tags.py` seems to be an unused template. There is some parsing for section titles and some templates in `pos.py`, but I don't think that's it. I could put it in `table.py` but some of those translations may be of use outside of tables (f.e. verb headers).

kristian-clausal:
I've been meaning to work on this, but there's been other things and it's been on the back-burner.

Stuff related to translating tags should all go into their separate files, especially a big dictionary of translations. If you can provide something like a spreadsheet with Greek -> English tags, that would be grand.

The issue I had was that Tatu really, really wants tags to be coded in a... not challenging, but annoying way. The English edition (not other editions) have a system for creating alternative tag sets that let you have different parallel tag collections for each thing being tagged; so if a form could be either `singular` or `plural` you would have two tag sets with `something singular` and `something plural`... And I always stall when trying to implement it. It doesn't help that there's a lot of messy, messy code in the English edition tag-related code bits that it becomes really easy to miss and forget things. Not to mention there's a few completely impenetrable algorithms in there that are used to parse strings and find the optimal matchups between raw strings and tags.

daxida:
I see what you mean. I only had in mind a simple version to start with. One not too intrusive as to not hinder any efforts towards what you describe. Do you think this is possible?

I have a couple questions regarding translations:
- How do hyphens exactly work? Is it supposed to be `first-person` or `first person`, `perfect-tenses` or `perfect tenses` etc.
- Capitalization: do I have to cover tags that only differ in their capitalization? Ex. ενεστώτας / Ενεστώτας > present. I see that the english tags are mostly lowercase but with exception like `Biblical`.

kristian-clausal:
I will implement the code, you can do translations of the contents of `raw_tags` in the current json output.

There are so many fiddly things to keep in mind, but we can start with:
- Only use valid tags present in `valid_tags` in `src/wiktextract/tags.py`. This is the canonical list of tags to aim for. We can enlarge it as needed. Some tags are auto-created, like capitalized tags.
[...]

PR #1403, opened 2025-10-10
[el] Merge θα particle in head parsing

daxida:
Just a minor bug fix that I found while testing #1394 

When parsing verb headers, the particle `θα` (rough equivalent of the English `will`), was treated as a separate form. This effectively meant that for a verb like `ψάχνω`, the json would be:

```json
    {
      "form": "θα", // will
      "raw_tags": [
        "στ.μέλλ"
      ]
    },
    {
      "form": "ψάξω", // search
      "raw_tags": [
        "στ.μέλλ"
      ]
    },
 ```

instead of the expected

```json
    {
      "form": "θα ψάξω", // will search
      "raw_tags": [
        "στ.μέλλ"
      ]
    },
 ```
 
I opted to merge the θα particle because it's consistent with the verb tables, but it is probably more useful to only keep the verb without the particles. At any rate, having a stray θα makes no sense.

It is a dirty ad-hoc hack, but then again, it is quite an edge case.

### DIFF
--- a/src/wiktextract/extractor/el/head.py
+++ b/src/wiktextract/extractor/el/head.py
@@ -1,8 +1,7 @@
 import re
+from unicodedata import name as unicode_name
 
 from mediawiki_langcodes import code_to_name
-
-from unicodedata import name as unicode_name
 
 from wiktextract.extractor.en.form_descriptions import distw
 from wiktextract.wxr_context import WiktextractContext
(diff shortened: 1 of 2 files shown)


kristian-clausal:
I suspect it's because the formatting is inconsistent on the wiktionary page.

> **ψάχνω**, *πρτ.*: **έψαχνα**, *στ.μέλλ.*: θα **ψάξω**, *αόρ.*: **έψαξα**, *παθ.φωνή*: **ψάχνομαι**, *π.αόρ.*: **ψάχτηκα**, *μτχ.π.π.*: **ψαγμένος**

The "tha" is not bolded. There's some complex and messy logic that handles the italics vs. unformatted vs. bolded stuff...

EDIT: There might be a way to handle this. I'll try to figure it out.

daxida:
That is correct. The θα is never bolded by the template. I should have probably stated it in the issue.

I went with the ad-hoc solution because tweaking other bits of logic seemed to me way harder (and easier to mess other parts).

PR #1405, opened 2025-10-10
[el] When partitioning head into formatted sections, if bold has been seen unformatted text is not form text

kristian-clausal:


kristian-clausal:
@daxida please check out the test to see if the output is right. The tags are now sorted by default so the outputs should not change.

kristian-clausal:
I just realized there's probably a way to push to the original PR, but eh.

daxida:
Seems like you decided to get rid of the θα altogether. That is fine with me too, but I don't know what's the standard, or what do other languages do in this matter. Keeping it is consistent with verb tables, but maybe verb tables should also get rid of them for all I know.

As long as there is no standalone θα that's an improvement in my book.

kristian-clausal:
There is no way to create a logic except a kludge for "should this unbolded bit of text be part of this following bolded piece of text", so I decided to err on this side. If God had wanted the "tha" to be in the form, he would have put it in bold.

daxida:
LGTM. 

The important part is the verb table anyway. This is just me wondering at why did a random θα appear in `forms`.

Issue #1409, opened 2025-10-13
[el] Parsing error in headers


daxida:
When parsing headers content, some entries have tag-related information appear in their form. F.e. [αιδώς](https://el.wiktionary.org/wiki/%CE%B1%CE%B9%CE%B4%CF%8E%CF%82)

```
wiktwords --db-path tmp/el_latest.db --edition el --language-code el --out outfile --page αιδώς
2025-10-13 08:49:50,247 INFO: Capturing words for: el
cat outfile | jq -c '{forms}' --indent 2
{
  "forms": [
    {
      "form": "αιδώς",
      "raw_tags": [
        "θηλυκό",
        "μόνο στον ενικό"
      ]
    },
    {
      "form": "λόγιο", // <-------- λόγιο := literary, this should not be here
      "raw_tags": [
        "θηλυκό",
        "μόνο στον ενικό"
      ]
    }
  ]
}
```

Parsing that information into a tag would be nice but probably the most important thing is to remove them as forms.

---

__How frequently does this appear?__

No idea. I had this snippet for diagnosing in `extractor/el/head.py`:

```py
            case ")":
                inside_parens = False
                # print(f"{current_forms=}, {current_tags=}, {t=}")
                if (
                    not current_forms
                    and len(current_tags) == 1
                    and code_to_name(current_tags[0]) != ""
                ):
                    # There are a lot of `(en)` language code tags that we
                    # don't care about because they're just repeating the
                    # language code of the word entry itself!
                    current_tags = []
                    continue

                # --------- ADDED THIS
                if not current_tags:
                    if current_forms == ["λόγιο"]:
                        # dirty hack: to remove λόγιο from forms
                        # TESTED: it does not affect λόγιο itself
                        current_forms = []
                        continue
                    with open("parens_cases.txt", "a") as f:
                        w = wxr.wtp.title
                        f.write(f"{w} -- {current_forms}\n")
                # ---------

                if current_forms and current_tags:
                    push_new_block()
                else:
                    extend_old_block()
```

And from the "parens_cases.txt" file (other examples of `λόγιο` can be found), more problems arise. F.e. [γάιδαρος](https://el.wiktionary.org/wiki/%CE%B3%CE%AC%CE%B9%CE%B4%CE%B1%CF%81%CE%BF%CF%82) 

```json
{
  "forms": [
    {
      "form": "γάιδαρος",
      "raw_tags": [
        "αρσενικό", // <---- masculine OK
[...]

PR #1410, opened 2025-10-13
[el] Improve head splitting and fixed bugs

kristian-clausal:
See #1409

Co-authored by daxida

https://en.wikipedia.org/wiki/Help:Colon_trick

Template:ετ was creating [[:Category:...]] links, which the code didn't handle.

Also fixed bugs with how the return value of
`wxr.wtp.node_to_text` was being unpacked, even though it was a string... Oops. Well, the contents would have been concatenated at some point anyhow, so there was no (afaict) actual changes in behavior and ultimate output.

Also handled HTML tag italics and bolds better (that is, at all).

Changed the logic when looking at the contents of a link: if there is bold text in the head, do not accept italics-only as a form.


daxida:
The indenting of the test is wrong. It is inside another test.

And when fixed, unfortunately, it does not pass.

But the test not passing may have been an oversight on my part. It did pass with the kludge mentioned in the issue, but some code changes have made it so that `page_datas` is now empty.

EDIT: Something wrong happened when copy pasting. I will PR your branch for the fixes.

EDIT2: Yeah, it was `raw` being indented that messed the input. Also I added `λόγιο` as `raw_tag` in expected.

PR #1412, opened 2025-10-13
[el] Fix logio test

daxida:


kristian-clausal:
Ah, that's what I get for not checking indent after copy-pasting... Thanks for checking this out.

PR #1416, opened 2025-10-14
[el] Support tables directly into the language level's content

daxida:
Solves this:

>         # XXX Some tables are put directly into the language level's content
>         # Separate content and sublevels, parse content and put in base_data

* Changes in `pos.py` are about calling `process_inflection_section` for `el-κλίση` template for tables at language level's content. We only support tables under `{{-el-}}` for this PR.
* Changes in `head.py` and `pos.py` are about not overwritting the language level's forms with header forms later down the line.
* Changes in `table.py` should make the code more similar to other extractors, and allow some room for (a minor) postprocessing of tags. (Tested)
* Changes in `tags.py` are just a bunch of wiki tags that I found while testing.
* Added an inflection test for some of the features (postprocessing, parse_inflection_table etc.).

### DIFF
--- a/src/wiktextract/extractor/el/pos.py
+++ b/src/wiktextract/extractor/el/pos.py
@@ -183,9 +183,11 @@ def process_pos(
         elif kind == NodeKind.ITALIC or (
             isinstance(node, HTMLNode)
             and (
-                node.tag == "span"
-                and "style" in node.attrs
-                and ("italic" in node.attrs["style"])
+                (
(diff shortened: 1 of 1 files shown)


kristian-clausal:
I've approved this, if you want to make any changes I'll not merge until you think it's ok.

daxida:
I added the fix to the if clause. I hope I got that right.

Other than that, I think it's ok.

PR #1423, opened 2025-10-15
[el] Populate Form `source` attribute

daxida:
- Populates Form `source` attributes with this string narrowing:
```py
FormSource = Literal[
    "conjugation",
    "declension",
    "header",
    "inflection",  # Can be further narrowed to conjugation/declension
    "linkage",
    "",
]
```
- Added a test with two form sources
- Fixed / moved tests to be easier to work with
- Some minor formatting changes and linting
- Add the `topics` attribute to every `Taggable`
- translate_raw_tags now populates topics.
- Added a bunch of tags/topics.

---
### Notes
- This works on a structural assumption:
  - Declensions **always** appear under **LEVEL2 Greek**.
  - Headers **always** are parsed via `parse_head` (reasonable).
  - Conjugations **always** appear anywhere else  
    *(this one is tricky, and I know some counter-examples, but all of them are editor's mistakes. I don't know about languages other than Greek though).*
- Went with keyword argument for `source` for clarity, but I'm fine with removing it.
- Went with lowercase (“declension” and not “Declension”), because in the English version I checked it was inconsistent (there was both “declension” and “Declension”), and the former seemed a better choice.

---
Edit:

Turns out I missed many forms that came from other places. I will change this to draft. There's still left:

- [x] Look at non Greek entries
  - Done partially, narrowed conjugation source to Greek entries.
- [x] Add Form `source` to other forms.
  - Done where I could find them while adding tags.
- [x] Further test the conjugation `source`

### DIFF
--- a/src/wiktextract/extractor/el/linkages.py
+++ b/src/wiktextract/extractor/el/linkages.py
@@ -193,7 +193,7 @@ def process_linkage_section(
                 for link_parts, ltags, _ in combined_line_data
             ]
             for form in transliteration_forms:
-                translate_raw_tags(form, wxr)
+                translate_raw_tags(form)
             data.forms.extend(transliteration_forms)
             if transliteration_template_data:
                 data.forms.extend(transliteration_template_data)
--- a/tests/test_el_tags.py
+++ b/tests/test_el_tags.py
@@ -1,20 +1,30 @@
 from unittest import TestCase
 
-from wiktextract.extractor.el.tags import base_tag_map
+from wiktextract.extractor.el.tags import tag_map, topic_map
 from wiktextract.tags import valid_tags
+from wiktextract.topics import valid_topics
 
 
 class TestElTags(TestCase):
     def test_validate_tags(self) -> None:
-        # Check validity of manually entered tags in base_tag_map.
+        # Check validity of manually entered tags in tag_map.
         # valid_tags is from higher level code, originally created for the English
         # extractor but also applicable to other extractors: these are the tags
         # that should be used for tagging. Can be added to when needed, but
         # often there's already an equivalent tag with a slightly different name.
-        for tags in base_tag_map.values():
+        for tags in tag_map.values():
             for tag in tags:
                 for part in tag.split("-"):
                     if not part.isalpha() or (
                         part.islower() and tag not in valid_tags
                     ):
                         self.assertFalse(f"Invalid tag in tag_map: {tag=}")
+
+    def test_validate_topics(self) -> None:
+        for tags in topic_map.values():
+            for tag in tags:
+                for part in tag.split("-"):
+                    if not part.isalpha() or (
+                        part.islower() and tag not in valid_topics
+                    ):
+                        self.assertFalse(f"Invalid topic in topic_map: {tag=}")


src/wiktextract/extractor/el/table.py
```diff
@@ -423,7 +430,12 @@ def parse_table(
                     # If a cell has no tags in a table, it's probably a note
                     # or something.
                     forms.extend(
-                        Form(form=text, raw_tags=list(tags)) for text in texts
+                        Form(
+                            form=text,
+                            raw_tags=list(tags),
```
kristian-clausal: `sorted(tags)` is tags is a set to keep it deterministic in case this needs to be in a test.

src/wiktextract/extractor/el/models.py
```diff
@@ -154,6 +156,9 @@ def merge(self, other: "Sense") -> None:
         self.related.extend(other.related)
 
 
+FormSource = Literal["declension", "conjugation", "header", ""]
```
kristian-clausal: Aaaah so this is how you're supposed to use `Literal`... I tried to figure them out a couple of years back, but apparently I didn't look deep enough. Of course it makes sense now, you need to type the variable... I thought you *had* to use an actual literal string or integer (as in `something("literal string used here")`)...

tests/test_el_inflection.py
```diff
@@ -46,28 +46,53 @@ def xinfl(self, word, pos, text):
...
+            if dontcare in received:
+                idx = received.index(dontcare)
+                received.pop(idx)
+
+    def mktest_form(self, raw, expected):
+        received = self.xinfl("filler", "verb", raw.strip())
+
+        self._remove_dontcares(received)
+
+        def normalize_forms(lst):
```
kristian-clausal: Let's not sort these in tests and instead have them sorted by the time the data comes to the test. The order should at least be deterministic. If a test fails because of this, fix the source by sorting a set with `sorted()` instead of `list()` etc.

daxida: If I understand correctly you want both tags and raw_tags to be sorted at every point of the program?

kristian-clausal: They should be deterministic, the output should be the same, not necessarily sorted-sorted. In practice it means that things need to be sorted only when using `set()` at some point, otherwise the order is deterministic. So things need to be sorted only when tests randomly break.

daxida:
Sorry for the many changes. One thing left to another while trying to diagnose if adding sources at some places made sense or not. The commits could have been cleaner.

I edited the original comment, but the main changes since your review are these:

> - Add the `topics` attribute to every `Taggable`
> - translate_raw_tags now populates topics.
> - Added a bunch of tags/topics.

To which you can include populating `source` at some more places.

The `tags.py` is still WIP (with a bunch of self-comments), but should work nicely as it is.

Moving it back to ready since I think the original objective of populating the `source` attribute should be done.

kristian-clausal:
Do not worry too-too much about the `source` field, it's nice to have, but not necessary... In fact, it's so not necessary that I've never used it for anything and had to ask Tatu about it, and he said that he added it just in case there's a good reason to separate data between different parts (tables mostly) that might vary in quality.

This all looks good to me. :+1:

daxida:
Honestly I was already doing something very similar to what `source` accomplishes via logging just to know where did those forms come from. Then I saw that there was an attribute for it and figured might as well use that.

Also it is true that the quality of the inflection tables is generally better. Wild things can happen in headers.

What I really wanted (and there's only a tiny work left) is to extract basic inflection information (gender, number etc.) from forms back to the word_entry itself. It's 5 or so lines, but to write those, I had to convince myself that they worked with all the above code.

kristian-clausal:
Please leave form-level tags in forms, unless you're 97% sure you can guarantee the data that bubbles up into the word entry tags is correct. I bet that even in Greek there are situations where you have overlap in such a way that one or the other of alternative ways to tag a word is not universally true for the whole word. You can kludge this and you can do it if it's specific, but not in general at least.

EDIT: Lowered sureness percent from 100% to 97%.

PR #1424, opened 2025-10-16
[fr] improve example list and conjugation table code

xxyzz:


xxyzz:
The table extractor code are written for the latest templates on Wiktionary. Many table header changes were made after Oct 1 and will be included in the 20251020 dump file.

PR #1427, opened 2025-10-19
[el] Add Greek tags

daxida:
### Main changes
- Added tag translations.
  - The coverage is 95% or so. I left some translations empty because they were no matching valid topics/tags. I left the comments and DONTCARES (usually parsing errors).
  - Warning about the mutable state of the English `valid_tags` / `valid_topics` that may yield surprising results.
- Fixed bug where Greek `raw_tags` (ex. θυληκό) could be found in `Sense` `tags` because they were not translated it in `convert_tags`.
  - Also for consistency, treat `no-gloss` always as a `tag`, which also makes so that we don't have to translate it.
- Fixed tests.
  
### Minor changes
- Linting
  - `copy` > `model_copy`, removing some unused imports / fixing others, and adding some type hints.
  - `mypy src/wiktextract/extractor/el/**py --strict` should pass now
- As a side note, I don't think `TypeAlias` is being used correctly most of the time but I didn't fix it since mypy seems fine with it.

### DIFF
--- a/tests/test_el_tags.py
+++ b/tests/test_el_tags.py
@@ -6,20 +6,6 @@ from wiktextract.topics import valid_topics
 
 
 class TestElTags(TestCase):
-    """WARNING:
-
-    The results differ when running these tests via unittest or manually here
-    as TestElTags().test_validate_tags. This is due to side effects adding more
-    tags/topics when running it via unittest.
-
-    As of the writting of this warning, there are (standalone/unittest):
-    - Tags   1315 / 3365
-    - Topics 227  / 854
-
-    This explains why some topics like "computer" pass these tests, even though
-    only "computing" is found in the original valid_topics.
-    """
-
     def test_validate_tags(self) -> None:
         # Check validity of manually entered tags in tag_map.
         # valid_tags is from higher level code, originally created for the English


kristian-clausal:
Please explain what is happening with test_el_tags.

daxida:
You can replicate those values themselves by adding

```py
print(len(valid_topics)) # on top, before the unittest `Class`.

# then run
python3 tests/test_el_tags.py # 227
python3 -m unittest tests/test_el_tags.py # 227 and FAILS TESTS
python3 -m unittest tests/test_el** # 854 and PASSES TESTS
```

For topics it's very simple, a `rg` search shows they are only modified at one place:

https://github.com/tatuylonen/wiktextract/blob/b042eb491dcf9d486dab85c1324a22d3c59ebdad/src/wiktextract/extractor/en/form_descriptions.py#L837

When running the testsuite, this line is called at some point before reaching test_el_tags.py, populates `valid_topics` and makes the test pass (because I used topics like `accounting` that passed the testsuite, without realizing that they were not in the original `valid_topics`...).

kristian-clausal:
Oh man, that code looks old... And I see a bug, it adds `tag` instead of `topic` to `sequences_with_slashes` on 839.

Topics are a bit ignored. When working with the EN extractor, I try to poke it as little as possible, because it mostly seems to be ok and people complain much, much, much less about issues with topics.

As for "accounting"... The code in `form_descriptions` is a bit messy. Yes, it adds stuff into `valid_topics`,  which I'm not sure it actually should do, because then it adds those same keys into the trie that is used to search for tags and topics...

EDIT: Ok, we've got tests that use this enlarged `valid_topics` selection, so I guess they're valid then.

daxida:
I have to admit that before writing the above comment, I was not aware that unittest could fail when calling that singular test.py file. I only knew that running it manually made it fail. That is not ideal, no matter if those topics are valid.

The way I see it, we can:
- Use some script to translate `topics` to `valid_topics` using that `topic_generalize_map` then...
  - Leave the warning. This should make the test pass when standalone, but will not prevent errors in the future. The warning should be enough when debugging. I'd rather do this.
  - Also cloning `_valid_tags_test = {*valid_tags}` in tags.py and use that in our tests. This should prevent errors in the future but adds some noise.
  - Fix `valid_topics` -- probably not an option

kristian-clausal:
I'm seeing if moving the code that added those tags and topics into `valid_tags` and `valid_topics` from `form_descriptions` into `tags` and `topics` works. That's the logical place where they should be, even if it's originally a kludge for the English edition.

When `form_descriptions` was still a top-level file (not in `extractor/en/`) this made a bit of sense, but not now when we want to decouple the `valid_` sets from the English extractor.

kristian-clausal:
Could you pull the branch in the above PR and see if that works out? Thanks.

daxida:
```
(wiktextract) rafa@rafapc ~/d/wiktextract (inflection-source)> python3 -m unittest tests/test_el_tags.py
854
3365
..
----------------------------------------------------------------------
Ran 2 tests in 0.000s

OK
```

It works. Those counts match the expanded version numbers of the WARNING.

The `test_el_tags` WARNING is no longer necessary. I would probably add in the english tags.py/topics.py that the set of `valid_tags` and `valid_topics` is expanded further down and contains other elements than those listed in the respective sets.

Also there are some suspicious set([words]) that could be rewritten as {words} in tags.py. Faster and one less indentation, but most likely no one cares.

Do you want me to add a commit removing the WARNING then?

kristian-clausal:
If you see anything that could be cleaned up like the `set()` thing, please feel free, lol. :+1:

Yeah, please remove the warning if it's not needed anymore.

When I merge this PR, it should also merge the other by association. Thanks!

daxida:
I think it's better if you merge yours and I rebase it from there. Otherwise the `sets` and comments will conflict with your changes.

At any rate, I'll commit the WARNING removal to begin with.

PR #1432, opened 2025-10-20
[lint] Update tags.py and topics.py set syntax

daxida:
Also comment about mutability of `valid_tags` and `valid_topics`.

Coming from a discussion in #1427

The github diff is awful...  I don't usually run the entire testsuite, but this is what I got:

```
(wiktextract) rafa@rafapc ~/d/wiktextract (tags-qol)> make test
python -m unittest discover -b -s tests

...

Ran 1593 tests in 30.394s

OK
```


kristian-clausal:
Oh THOSE `set([])`. Man, I didn't even notice those, I just trusted there were some... Thanks for checking this out.

PR #1437, opened 2025-10-21
[el] Expand table form variants

daxida:
Basically expands some table forms when they have a `(some_text)` in them. Example from [πίνω](https://el.wiktionary.org/wiki/πίνω): `πίνουν(ε)` into `πίνουνε` AND `πίνουν`

This is coherent with the English version:

---

https://kaikki.org/dictionary/All%20languages%20combined/meaning/%CF%80/%CF%80%CE%AF/%CF%80%CE%AF%CE%BD%CF%89.html
image

---

https://en.wiktionary.org/wiki/%CF%80%CE%AF%CE%BD%CF%89#Greek
image

---

Where you can see that πίνουνε/πίνουν are treated as two forms, even though πίνουνε never appears in the wiki page.

This may seem to conflict with our conversion of parens when meaning *rare* for declension, but those should always have been dealt with before form expansion, since they *contain* the word (f.e. (μπογιών)), while these variants are *inside* it (f.e. πίνουν(ε)). One could always fix the regex to match at least one character *before* if this becomes an issue (I don't think it will).

### DIFF
--- a/tests/test_el_inflection.py
+++ b/tests/test_el_inflection.py
@@ -604,10 +604,10 @@ class TestElInflection(TestCase):
         self.mktest_form_no_raw_tags(raw, expected)
 
     def test_el_conjugation_table(self) -> None:
-        # Section of https://el.wiktionary.org/wiki/πίνω:
+        # Section of https://el.wiktionary.org/wiki/πίνω
         # {{el-κλίσ-'μπαίνω'|θαορ=ήπι|θμελλ=πι}}
         # Expanded via 'wxr.wtp.node_to_text(node)' at the start of
-        # 'process_inflection_section'
+        # 'table.process_inflection_section'
         raw = """
         {|
         |-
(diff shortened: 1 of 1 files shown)


src/wiktextract/extractor/el/table.py
```diff
@@ -473,6 +477,31 @@ def postprocess_table_forms(forms: list[Form]) -> None:
...
+    # >
+    # { "form": "πίνουν", ...rest }
+    # { "form": "πίνουνε", ...rest }
+    new_forms: list[Form] = []
+    for form in forms:
+        text = form.form
+        if not text:
+            continue
+
+        m = re.match(r"^(.*?)\((.*?)\)(.*)$", text)
```
kristian-clausal: Does the job, but I would bet you there are other parenthesized things out there that will mess with it (like maybe "form (rare form)" which isn't caught by the bit that tags rare words)... Because there always is :smiling_face_with_tear: . Good enough for now. :+1:

kristian-clausal (approved):

Issue #1398, opened 2025-10-08
[en] Weirdness when parsing derived/related terms in tables and columned lists


Ian-07:
Several entries have multiple derived/related terms incorrectly combined into a single item. These are all inside {{col}} templates, and the issue seems to occur when there are second-level bullet points involved. See [pseph-](https://kaikki.org/dictionary/All%20languages%20combined/meaning/p/ps/pseph-.html) for example:

```
"derived": [
    {
      "topics": [
        "geology",
        "geography",
        "natural-sciences",
        "geology",
        "geography",
        "natural-sciences"
      ],
      "word": "psephitepsephitic"
    },
    {
      "tags": [
        "historical"
      ],
      "word": "psephomancy"
    },
    {
      "word": "psephocracy"
    },
    {
      "tags": [
        "rare"
      ],
      "word": "psephograph"
    },
    {
      "word": "psephologypsephologicalpsephologicallypsephologist"
    }
  ],
```

Some more examples: [coen-](https://kaikki.org/dictionary/All%20languages%20combined/meaning/c/co/coen-.html), [dendr-](https://kaikki.org/dictionary/All%20languages%20combined/meaning/d/de/dendr-.html), [homosexual](https://kaikki.org/dictionary/All%20languages%20combined/meaning/h/ho/homosexual.html), [-logia](https://kaikki.org/dictionary/All%20languages%20combined/meaning/-/-l/-logia.html), [pathology](https://kaikki.org/dictionary/All%20languages%20combined/meaning/p/pa/pathology.html)

A similar issue appears on entries which transclude [Template:table:Solar System/en](https://en.wiktionary.org/wiki/Template:table:Solar_System/en), e.g. [Earth](https://kaikki.org/dictionary/All%20languages%20combined/meaning/E/Ea/Earth.html):

```
    {
      "word": "PhobosDeimos"
    },
    {
      "word": "IoEuropaGanymedeCallisto"
    },
    {
      "word": "MimasEnceladusTethysDioneRheaTitanIapetus"
    },
    {
      "word": "MirandaArielUmbrielTitaniaOberon"
    },
    {
      "word": "Triton"
    },
    {
      "word": "Charon | style=\"width: 7.1%"
    },
    {
      "word": "white-space: nowrap"
    },
    {
      "word": "\" | Dysnomia"
    },
```

kristian-clausal:
[...]

PR #1440, opened 2025-10-22
[en] When checking linkage sections, recurse through HTML lists

kristian-clausal:
Also see tatuylonen/wikitextprocessor#401 that adds `is_list()` and `is_list_item()` functions.

Part of the issue in #1398 was that `Template:col` was not pre-expanded (all the other aliases of col were, though??), but even when pre-expanding there was an issue because the expanded template outputted HTML lists instead of wikitext.

It is correct to also recurse through HTML lists and list items, and fixing it did separate the different lines so that the output was not "concatenatedtogetherlikethis", but the way linkage items are parsed assumes that the hierarchy is like this:

```wikitext
* something that should be a tag:
** example forms
```

NOT:

```wikitext
* an example form
** an example form with a suffix
*** a third example form derived from the item directly above
```

This needs checking.

Do not merge until the PR in wikitextprocessor has been merged.

### DIFF
--- a/src/wiktextract/extractor/en/page.py
+++ b/src/wiktextract/extractor/en/page.py
@@ -27,6 +27,8 @@ from wikitextprocessor.parser import (
     NodeKind,
     TemplateNode,
     WikiNode,
+    is_list,
+    is_list_item,
 )
 
 from ...clean import clean_template_args, clean_value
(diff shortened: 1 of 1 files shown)


kristian-clausal:
Test fail due to the dependency, wikitextprocessor doesn't have `is_list()` yet.

PR #1442, opened 2025-10-23
[el] Support diminutives/augmentative form_of

daxida:
- Supports four templates for diminutives/augmentative  forms
- Does not add tags, since they end up being there anyway via `{{υπο}} > raw_tag: υποκοριστικό > (translation) > tag: diminutive`
- Does not issue a wiki_notice since not having a lemma (first template argument) is an intended use, documented in the wiki template page. In that case it does nothing (no `1 in t_node.template_parameters`).


kristian-clausal:
Thanks for your contribution again!

PR #1445, opened 2025-10-27
[el] `Postprocess_table_forms` bugfixes

daxida:
* Also remove `"ο/η", "του/της", "τον/τη", "τους/τις"` when removing table articles. Note that the logic for this is a bit different than the remaining `ARTICLES` because `parse_table` already separated them into a different form.
* Ignore suffix/prefixes (-ισμός) when separating (introduced by #1441)
* Trim trailing numbers.


kristian-clausal:
Yeah, this LGTM, thanks for the contribution!

PR #1449, opened 2025-10-29
[el] Translate tags in `Linkage`

daxida:
Title. Also a couple of tags added.

Stems from [πηγάινω](https://el.wiktionary.org/wiki/πηγαίνω), where I could not find this section in the jsonl:

image


kristian-clausal:
Thanks, looks good to me!

PR #1451, opened 2025-10-30
[el] Use POSName type at more places - refactor

daxida:
* Rename `type` to `heading_type` to not use the reserved keyword and for consistency.
* Remove `heading_name` since identical to `heading_title`. `parse_lower_heading` took a `heading: str` as argument that was always returned. Use only `heading_title`.
* Use `POSName` where a `pos` is expected. Note that since `POSName` is just a string, this is nothing but documentation via types.

Stems from an intent to narrow `POSName` to a Literal union. It is quite hard to do without an ADT since we have to convince mypy that, in many functions of `parse_utils.py`: `Heading.POS` => `(POSName | SectionName)` is actually `POSName`. This PR doesn't achieve that, `POSName` remains a `str` but the changes should hopefully make things easier.

### DIFF
--- a/src/wiktextract/extractor/el/parse_utils.py
+++ b/src/wiktextract/extractor/el/parse_utils.py
@@ -1,6 +1,6 @@
 import re
 import unicodedata
-from typing import Generator
+from typing import Generator, TypeAlias
 
 from wikitextprocessor import LevelNode, WikiNode
 
@@ -78,9 +78,11 @@ GREEK_LANGCODES = set(
 )
 
 
-Title = str
-SectionName = str  # The keys of SUBSECTION_HEADINGS
-POSReturns = list[tuple[POSName, Title, Tags, int, WikiNode, WordEntry]]
+Title: TypeAlias = str
+SectionName: TypeAlias = str  # The keys of SUBSECTION_HEADINGS
+POSReturns: TypeAlias = list[
+    tuple[POSName, Title, Tags, int, WikiNode, WordEntry]
+]
 
 
 def find_sections(


src/wiktextract/extractor/el/parse_utils.py
```diff
@@ -78,56 +78,59 @@
 )
 
 
-Title: TypeAlias = str
-
-POSReturns: TypeAlias = list[
-    tuple[POSName, Title, Tags, int, WikiNode, WordEntry]
-]
+Title = str
```
kristian-clausal: Please annotate these with TypeAlias. It is not strictly necessary, but I want them to be explicitly marked as being type aliases. If these should not be annotated as TypeAlias, what is TypeAlias for?

src/wiktextract/extractor/el/parse_utils.py
```diff
@@ -78,56 +78,59 @@
...
+                f"Sub-sub-section is numbered: {heading_title}, {num=}",
                 sortid="page/find_pos_sections_1",
             )
-        yield type, pos, heading_name, tags, num, node
+        yield heading_type, pos_or_section, heading_title, tags, num, node
 
 
 def parse_lower_heading(
     wxr: WiktextractContext, heading: str
-) -> tuple[Heading, str, str, Tags, int, bool]:
```
kristian-clausal: Yeah, removing the second `str` here is probably correct. I was probably thinking of doing something more exciting with it other than just returning it...

src/wiktextract/extractor/el/parse_utils.py
daxida: Also:
* `Title: TypeAlias = str` but `POSName = str`.
* `POSReturns: TypeAlias = list[etc.]` but `Tags = list[str]`

kristian-clausal:

PR #1452, opened 2025-10-30
[el] Narrow `POSName`

daxida:
* Narrow `POSName` with a literal
* Please mypy by casting at some places. Explain the rationale.

Note that this is more that type hints. Pydantic validates `POSName`, since it appears as the type of `pos` in `WordEntry`. This effectively mean that the code will crash if something not in `POSname` is assigned to `wordentry.pos`. It also means that we are guaranteed that only strings in that list will appear on the final jsonl.


kristian-clausal:
Thanks, LGTM!

PR #1453, opened 2025-10-30
[el] Expand adj/participle forms, remove articles in head

daxida:
- Expand adjectives and participles headword forms. That is, `άμεσος, -η, -ο` > ["άμεσος", "άμεση", "άμεσο"]. This also populates tags for those 3 forms. Note that this requires the `pos_tags` that we never used in `process_pos` to be added to `data` so that we only expand when it is an adjective or a participle (it is the participle that requires tags).
- Remove articles in headwords (logic shared with table forms)
- Trim trailing stars in forms.


kristian-clausal:

PR #1455, opened 2025-10-31
[el] Add alternative templates for syn/antonyms inside βλ

daxida:
The only real changes are a couple lines in `pos.py`. 

The rest is the adding of a test (`test_synonyms_antonyms_alt_templates`) and some cleaning.


kristian-clausal:
Thanks!

PR #1456, opened 2025-11-02
[el] Various fixes

daxida:
- Also remove `τον/την` from articles
- Fix a couple verb table tags, and sync tags in the two most common tables. This comes with a test.
- Refactor `extract_form_of_templates_ptosi` to be more in line with other extractions. Instead of warning, only call the function when there is `1` in template_parameters


kristian-clausal:
Thanks for the contribution!

PR #1457, opened 2025-11-03
[el] Narrow `LinkageType`

daxida:
Notes:
- I really didn't want to, but I had to change the `ifs` to a `match` to please mypy in `pos.py`
- Renames `type` to `linkage_type` to match `process_linkage_section` arg.

This is in draft because:
- Solved. ~~Type analysis now points that the `Heading.Derived` branch is unreachable, and indeed it is never passed to `process_linkage_section`. Is this intended?~~
- While `test_transliterations_esperanto` correctly seems to be a transliteration, most of the time I've seen this `άλλη γραφή` (expanded to `άλλες γραφές`) heading used for Greek, and how it is documented in the associated [γρ](https://el.wiktionary.org/wiki/Πρότυπο:γρ) template, is for alternative spellings. Think civilise, civilize in English. 
Collecting them as forms, which is what happens now, seems the right approach to me. I would get rid of the `transliteration` tag though, and add the extremely close `άλλες μορφές` heading to the same category. Maybe a `Heading.AltForm` variant?
- Expressions (Εκφράσεις) are not properly parsed in `Linkage`. Cf. [this](https://el.wiktionary.org/wiki/ψύλλος) or [this](https://el.wiktionary.org/wiki/κώλος) - Does the `en` version also group them in `Related`?

### DIFF
--- a/src/wiktextract/extractor/el/linkages.py
+++ b/src/wiktextract/extractor/el/linkages.py
@@ -45,14 +45,16 @@ def process_linkage_section(
             # print("REACHED")
             # print(f"{node.largs=}")
             ret: list[Node] = []
-            # print(f"{ret=}")
             comma = False
-            for arg in node.largs[1:]:
+            for k, v in node.template_parameters.items():
+                if not isinstance(k, int):
+                    continue
                 if comma:
                     ret.append(", ")
-                ret.append("__L__")
-                ret.append(wxr.wtp.node_to_text(arg))
-                ret.append("__/L__")
+                if isinstance(v, list):
+                    ret.extend(["__L__", *v, "__/L__"])
+                else:
+                    ret.extend(["__L__", v, "__/L__"])
                 comma = True
             return ret
         if node.template_name in ("eo-h", "eo-x"):
--- a/tests/test_el_glosses.py
+++ b/tests/test_el_glosses.py
@@ -6,6 +6,7 @@ from wikitextprocessor import Wtp
 from wiktextract.config import WiktionaryConfig
 from wiktextract.extractor.el.models import WordEntry
 from wiktextract.extractor.el.pos import process_pos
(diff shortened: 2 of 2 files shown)


src/wiktextract/extractor/el/linkages.py
```diff
@@ -180,6 +189,7 @@ def links_node_fn(
         case Heading.Antonyms:
             target_field = data.antonyms
         case Heading.Derived:
+            # unreachable
```
kristian-clausal: Isn't this unreachable because you've narrowed down linkage_type by using LinkateType, which does not have Heading.Derived?

daxida: If it is future-proofed as you say, then it's all good. I was afraid it was intended to be there (in the function), but missing in the if condition above. That is, that it should have been:

```py
        elif heading_type in (
            Heading.Related,
            Heading.Synonyms,
            Heading.Antonyms,
            Heading.Transliterations,
            Heading.Derived,
        ):
            process_linkage_section(wxr, data, sl, heading_type)
   ```

Comment (kristian-clausal) [reply src/wiktextract/extractor/el/linkages.py]: Just remove it then.

**Comment by daxida** [comment]
Added a bugfix to prevent:

```
===={{συνώνυμα}}====
* {{βλ|0=-}} [[πισινός#Ουσιαστικό|πισινός]]
```

to be parsed as `"synonyms": [{"word": "0=- πισινός"}]`

I will put this as ready, because, while I tackled the second point, fixing it properly by using `alt_of` is not supported by the `kaikki-to-yomitan` repo that I use, and I don't want to make my life harder (...for now).

kristian-clausal:
This looks good to me, thanks! EDIT: I did check it out earlier when seeing the e-mail about the commit, before the above comment, if someone is wondering.

PR #1462, opened 2025-11-06
[el] Minor fixes

daxida:
* Remove `τη(ν)` from article forms
* Test validity of heading tags, and prune the only offender.


kristian-clausal:
Thanks!

PR #1464, opened 2025-11-07
[el] Remove dead code

daxida:
By no means exhaustive, just the low hanging fruit.

Simple tags was not used, and the comments in `pos.py` are already addressed/documented somewhere else. 

Note: 
* I edited the [θηλυκός](https://el.wiktionary.org/wiki/θηλυκός) wiki because the και is not standard in head forms. It should have been like [this](https://el.wiktionary.org/wiki/μαλακός): `-ή/-ιά, -ό`
* I translated άκλιτος by `invariable` back then, not `uninflected`.  Looking at [the english equivalent page](https://en.wiktionary.org/wiki/καφέ) they go with `indeclinable` but, when clicking the tag, it lists `invariable` as an alternative.


kristian-clausal:
LGTM, I have branches slowly percolating on the backburner that I haven't pushed that might use simple_tags, but I can just re-add that stuff if needed later.

PR #1472, opened 2025-11-10
[de] fix for "Englisch Verb Übersicht" table template

xxyzz:
stubborn admins resist editing their precious templates


kristian-clausal:
> stubborn admins resist editing their precious templates

I couldn't handle it, I gave up on doing stuff on the wiki-side.

xxyzz:
The reasons(more like excuses) they made were total bs...

PR #1473, opened 2025-11-10
[doc] Typos and suggestions

daxida:
Just a couple suggestions over #1471, but more than anything, an excuse to ask if you plan on dropping python 3.10 support any time soon?

I am not sure the English version uses those `TypedDicts`, but having the `Required` type hint for `TypedDicts` (3.11>=) just makes them much more robust. I know that, at least, it will silence quite some IDE warnings for the Greek edition.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ You can also easily pretty-print the data into a more human-readable form using:
 print(json.dumps(data, indent=2, sort_keys=True, ensure_ascii=False))
 ```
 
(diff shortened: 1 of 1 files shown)


xxyzz:
I think we could drop 3.10 next year when it's end of security fixes unless we have to use features in 3.11. Typing is not that important there are other things with higher priority... And pydantic is the proper tool for enforcing types.

But I think drop 3.10 now is also fine, we'll have 4 versions supported after dropping 3.10. I'd like to wait for Kristian's opinion.

kristian-clausal:
As usual, let's leave changing Python versions as late as possible, unless there's something super juicy. `Required` isn't imperative, we will survive for a while.

PR #1489, opened 2025-11-18
[el] Fix participle pos

daxida:
Not sure if this needs a Greek grammar explanation. Please tell me it does not.

Note that the English version does label ζαρωμένος as adjective. Then again, it is just under an adjective template there, which unsurprisingly helps the matter... [el](https://el.wiktionary.org/wiki/%CE%B6%CE%B1%CF%81%CF%89%CE%BC%CE%AD%CE%BD%CE%BF%CF%82#Greek) [en](https://en.wiktionary.org/wiki/%CE%B6%CE%B1%CF%81%CF%89%CE%BC%CE%AD%CE%BD%CE%BF%CF%82#Greek)

---

Also, this does not type check. It requires this overload in `wikitextprocessor` that I am too lazy to send in a PR. If you want to add it yourself, feel free:

```py
    @overload
    def find_child_recursively(
        self,
        target_kinds: Literal[NodeKind.TEMPLATE],
    ) -> Iterator["TemplateNode"]: ...
 ```

### DIFF
--- a/src/wiktextract/extractor/el/page.py
+++ b/src/wiktextract/extractor/el/page.py
@@ -155,6 +155,16 @@ def parse_page(
                         source="declension",
                     )
 
+        # If we only looked at the headword (μετοχή), all the participles would
+        # be labeled as "verb"s, which is just wrong.
+        # Instead, we quickly inspect the page for a participle template, and
+        # then, if we are indeed a participle, we update the pos.
+        better_participle_pos: POSName | None = None
+        for child in page_root.find_child_recursively(NodeKind.TEMPLATE):
+            t_name = child.template_name
+            if t_name.startswith("μτχ") and t_name != "μτχεε":
+                better_participle_pos = "adj"
+
         for sublevel in sublevels:
             if len(sublevel.largs) == 0:
                 wxr.wtp.debug(
@@ -169,6 +179,14 @@ def parse_page(
             heading_type, pos, tags, num, ok = parse_lower_heading(
                 wxr, heading_title
             )
+
+            # It would have been great if we had a "participle" pos, but alas!
+            if (
+                pos == "verb"
+                and tags == ["participle"]
+                and better_participle_pos
+            ):
+                pos = better_participle_pos
 
             section_num = num if num > section_num else section_num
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by daxida** [comment]
Btw. I think that just having a "participle" pos solves all these issues and many more, but it does not seem to be a thing in the English version.

**Comment by kristian-clausal** [comment]
We would rather not "correct" wiktionary too much, just have the data as is. If you feel strongly about it, feel free to create a new `participle` POS; the contents of the POS field is not that strictly restricted to just the things found in the original extractor.

**Comment by daxida** [comment]
> We would rather not "correct" wiktionary too much, just have the data as is. If you feel strongly about it, create a new participle POS; the contents of the POS class is not that strictly restricted to just the things found in the original extractor.

I don't think it's as much correcting wiktionary as correcting the claim that participles **are verbs** made here by the English extractor, where one can read, in `PARTS_OF_SPEECH`:

```python
    "participle": {
        "pos": "verb",
        "tags": ["participle"],
    },
```

[Wikipedia](https://en.wikipedia.org/wiki/Participle) is more cautious:

> In linguistics, a participle (from [Latin](https://en.wikipedia.org/wiki/Latin_language) [participium](https://en.wiktionary.org/wiki/participium#Latin) 'a sharing, partaking'; abbr. PTCP) is a [nonfinite verb form](https://en.wikipedia.org/wiki/Nonfinite_verb) that has some of the characteristics and functions of both verbs and adjectives.[[1]](https://en.wikipedia.org/wiki/Participle#cite_note-1) More narrowly, participle has been defined as "a word derived from a verb and used as an adjective, as in a laughing face".[[2]](https://en.wikipedia.org/wiki/Participle#cite_note-2)

_derived from a verb and used as an adjective_

As for Greek, wiktionary has templates from adjectives appended to these forms. For ζαρωμένος, it is `{{el-κλίση-'όμορφος'}}`. It makes no sense to me to claim that a `verb` can be masculine. And **here**, if I had to decide on a pos between `verb` and `adj`, adjective is the better choice by far.

Now, I agree that this is not the general case: participles can be closer to verbs at times (or even to adverbs in Greek, or to nouns in Spanish, from the top of my head), so my question is why enter this discussion to begin with and not just label the pos `participle`.

I will close this PR anyway because I think it is too late to change it everywhere, and as long as it is consistent, I guess I can always check if the pos is a verb AND there is a participle tag to know if it is a participle in every extracted language.

PR #1495, opened 2025-11-20
[el] Remove more articles from forms

daxida:
Self explanatory. The test is overkill, but that's how I found the issue.


xxyzz:
Two suggestions:
* If article data always in odd index number column, maybe there is not need to hard code them. It's even more easier to edit the template to add a unique class name or something to these columns if the admins agree with the change.
* Test wikitext doesn't need to be escaped and the expanded wikitext can be obtained from [Special:ExpandTemplates](https://el.wiktionary.org/wiki/Ειδικό:ΕπέκτασηΠροτύπων) page. This is more readable and if CSS rules are not used by extractor code they can be removed.

daxida:
* The current postprocess applies to every table template (κλίση for nouns, κλίσ for verbs). Your suggestion is probably better than what is done now, but would require to discriminate because verbs most likely have relevant information at those positions. Also I think with this extra "article form" should be the last (one can hope!).
* That's handy, thank you for letting me know about it.

For the trivia, it's easy for me to detect these, because the `kty` dictionary will redirect those forms when hovering articles (since they are considered forms or whatever word they appear in, if not trimmed), making the dictionary lag enough to be noticeable :D

image

PR #1501, opened 2025-11-24
[el] Don't add forms if there is no template name

daxida:
Alternatively, you could close this and consider fixing whatever it is that happens at [κρόκος](https://el.wiktionary.org/wiki/%CE%BA%CF%81%CF%8C%CE%BA%CE%BF%CF%82) with the pictures.

Without this you will find:

```json
{"tags":["inflection-template"],"source":"inflection"},{"form":"thumb|200px|Ο κρόκος ενός αβγού.","raw_tags":["thumb|150px|Άνθη κρόκου Crocus sativus"],"source":"inflection"}
```

in the forms.

### DIFF
--- a/src/wiktextract/extractor/el/table.py
+++ b/src/wiktextract/extractor/el/table.py
@@ -105,6 +105,7 @@ def process_inflection_section(
     snode: WikiNode,
     *,
     source: FormSource = "",
+    top_template_name: str | None = None
 ) -> None:
     table_nodes: list[tuple[str | None, WikiNode]] = []
     # template_depth is used as a nonlocal variable in bold_node_handler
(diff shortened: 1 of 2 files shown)


xxyzz:
IMO, adding template name to "forms" list as a form feel strange and it probably should be added to `Form.source` field.

kristian-clausal:
> IMO, adding template name to "forms" list as a form feel strange and it probably should be added to `Form.source` field.

This makes sense. Do you think there's anyone who is currently relying on the old behavior?

xxyzz:
I can't imagine how this template name data could be useful to most people... even if they use it add it to `Form` object makes clearer which words are extracted from the template.

daxida:
Not entirely sure if "relying", but `kaikki-to-yomitan` blacklists this tag. That is, if a form contains this tag, it is, understandably, not ending up in the dictionary.

If you were to delete it nothing will happen, other than they will be checking for a non-existent tag in order to filter it out.

I have to add that knowing the template name is sort of useful when debugging, but yeah, maybe not having it as a Form is the right call.

PR #1507, opened 2025-11-25
[it] should check lower case headword line template name

xxyzz:


xxyzz:
I only realized GitHub doesn't support fast-forward merge today... this branch was fast-forward merged locally than pushed to the master branch. The previous pull requests merged today were merged with the "sqush and merge" button on GitHub. But of course use the "merge" button is fine, I just don't understand why they don't support it.

PR #1531, opened 2025-12-03
Add a script to compare json schemas

daxida:
It looks like this:

image
image

It shows the languages on hover.

I refactored `generate_schema` so that I could use the iteration logic instead of reading `_site` (which I was doing before). It should work the same.

The output folder you may change it to whatever, in case you decide to add it (or not) to the website with the schemas.

Also there are some fields only present in the Simple extractor. Those you probably want to get rid of since they are not used anywhere else.

### DIFF
--- a/tools/compare_schemas.py
+++ b/tools/compare_schemas.py
@@ -1,0 +1,228 @@
+"""Generate a simple html page to visualize which attributes are in which
+language.
+
+Internally, it just compares the json schemas made from Pydantic (so no English
+edition!).
+"""
+
+import json
+from pathlib import Path
+from typing import Any, TypedDict
+
+from generate_schema import iter_schemas
+
+Attrs = list[str]
+
+
+class Summary(TypedDict):
+    language: str
+    attributes: Attrs
+
+
+def summarize_schema(schema: Any) -> Summary:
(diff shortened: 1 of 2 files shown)


kristian-clausal:
I originally mentioned that this should go in `usertools`, not `tools`, but looking at what's in them it's a mess, and this fits either.

`tools` should be for stuff that generates things that we actually later copy or use in stuff that goes to `src` (generating dicts, for example), while usertools is for analyzing data in more general ways from a user perspective. This straddles that, so it fits `tools` as well.

daxida:
Oops, I only put it there because it was using the `generate_schema` script from that folder. Unfortunate.

xxyzz:
Thanks! This is quite helpful to check all schema fields. I made some changes and added it to GitHub page: https://tatuylonen.github.io/wiktextract/compare_schemas.html

daxida:
Happy to hear that. I saw that you fixed it (read: made my code less awful), so thank you.

I wanted to change a couple things about the Greek edition using that, but got sidetracked by this German thingy.

Also I find the website sort of burried in the README, but then again, I'm probably in the minority of users that are actually interested in the schemas -- and I can always click the `Deployements` icon on top to be redirected to a link.

PR #1539, opened 2025-12-05
[el] Fix wrong form expansion in hyphenated words

daxida:
Kristian warned me about it, but since these are so rare I couldn't get my hands on any concrete example.

We now just ignore separator expansion if the original word contained it (mainly for hyphens, I don't think there are words with /)


kristian-clausal:
Yeah, there's always, always something. Thanks for the PR!

Issue #1544, opened 2025-12-05
Quick question about declension tables


daxida:
What is the official stance about declension tables containing articles?

I checked https://github.com/tatuylonen/wiktextract/issues/161 and the docs but I could not find a clear answer.

They are trimmed in the Greek edition, and from a quick browsing of the English edition, most declension tables don't have articles [aqua](https://en.wiktionary.org/wiki/aqua#Latin) [κόκκινο](https://en.wiktionary.org/wiki/κόκκινο#Greek) so there is no problem there. The issue is with German, where the articles are still there, and cause major issues when making dictionaries. Of course, this is not a wiktextract issue properly speaking and you may as well not care.

And if there is a definite answer, could it be added somewhere in the docs? I know it's supposed to be for creating extractors but I think it's a great resource for clearing up doubts about conventions. I would suggest doing the same with `tags` and `raw_tags` from https://github.com/tatuylonen/wiktextract/issues/1418 (should tags be sorted, not sorted, don't care etc.), but I guess if there is no consensus then it's tricky.

xxyzz:
I think pronoun and article data should be separated from the form words if they could be detected easily and reliably. For example, in German Wiktionary's [Deutsch Verb Übersicht](https://de.wiktionary.org/wiki/Vorlage:Deutsch_Verb_%C3%9Cbersicht) table template, there is a column of pronoun data and they are extracted to the "Form.pronouns" list. But most other table templates put both article and form word in the same column and also in plain text therefore they are not separated.

xxyzz:
Article data in [Deutsch Substantiv Übersicht](https://de.wiktionary.org/wiki/Vorlage:Deutsch_Substantiv_%C3%9Cbersicht) template added to `Form.article`: #1545

kristian-clausal:
If the articles are simple to detect and remove, they should be, unless there's some linguistic reason why they shouldn't -- for example, if the articles can be irregular, as in, you can't predict what the article's form is based on the other information found in the data. This goes roughly for other particles and clitics and such; affixes without spaces would be too bothersome to remove, so those can be left as is.

daxida:
I personally have no issues but I just wanted to add, in case the result is surprising for some users, that by separating the articles some grammatic information is lost in the process.

This is coherent, and follows the logic of other editions of not adding a `Form` if `Form.form` is equal to `WordEntry.word`.

[...]

PR #1549, opened 2025-12-08
[de] Extract articles from a couple other templates

daxida:
Expand https://github.com/tatuylonen/wiktextract/pull/1545 for a `name` template and the `adjective` in the testsuite.

Note that I had to move the `separate` function to flexion.py because inflection.py imports it, so there was a circular dependency if I wanted to use it in both files.

Cf. also https://github.com/tatuylonen/wiktextract/issues/1544


daxida:
Actually, if this is well received, consider not merging it and I will add more templates to this PR. Let me know what you think.

xxyzz:
Thanks!

Issue #1498, opened 2025-11-23
[en] Japanese template not being parsed


daxida:
I've seen this happen in a couple pages. Take this for instance: おやこ
[wiktionary](https://en.wiktionary.org/wiki/おやこ#Japanese)
[kaikki](https://kaikki.org/dictionary/Japanese/meaning/%E3%81%8A/%E3%81%8A%E3%82%84/%E3%81%8A%E3%82%84%E3%81%93.html)

The `Alternative Spelling` table on the top right, corresponding to:

```
==Japanese==
{{ja-kanjitab|alt=親子,母子,母娘,父子,父娘}}
```

does not seem to be parsed anywhere.

Note that it can also happen inside etymology (where it makes more sense and should be): きょう
[wiktionary](https://en.wiktionary.org/wiki/きょう#Japanese)
[kaikki](https://kaikki.org/dictionary/Japanese/meaning/%E3%81%8D/%E3%81%8D%E3%82%87/%E3%81%8D%E3%82%87%E3%81%86.html)

```
===Etymology 1===
{{ja-kanjitab|alt=今日}}
```

but even then, I can't find it in `etymology_templates`

---

For self reference, you may ignore this. Maybe related to https://github.com/yomidevs/kaikki-to-yomitan/issues/275

xxyzz:
"*_templates" fields are unreliable because pre-expanded templates and some table templates are not added.

daxida:
> "*_templates" fields are unreliable because pre-expanded templates and some table templates are not added.

Are the unreliable fields documented? I only tried to avoid `raw_tags` and `raw_glosses` but now that I know about templates, it makes me wonder what else may be there.

kristian-clausal:
> > "*_templates" fields are unreliable because pre-expanded templates and some table templates are not added.
> 
> Are the unreliable fields documented? I only tried to avoid `raw_tags` and `raw_glosses` but now that I know about templates, it makes me wonder what else may be there.

Unfortunately, owing to the nature of the code and the project, I would consider all fields "unreliable".

[...]

PR #1552, opened 2025-12-09
[ja] Postprocess ja conjugation tables

daxida:
Cf. https://github.com/tatuylonen/wiktextract/issues/1498#issuecomment-3569385805

I first commited the whole table so you can compare the commits with and without the postprocessing.

Loe and behold the worst mouse writing ever:

2025_12_09_0cv_Kleki

`table_name` became unused (because it's always table1/table2, with no relevant information), and could be removed entirely, but I'd like to know what you think first.

### DIFF
--- a/src/wiktextract/extractor/ja/conjugation.py
+++ b/src/wiktextract/extractor/ja/conjugation.py
@@ -1,4 +1,6 @@
-from wikitextprocessor import LevelNode, NodeKind, TemplateNode
+import re
+
+from wikitextprocessor import LevelNode, NodeKind, TemplateNode, WikiNode
 
 from ...page import clean_node
 from ...wxr_context import WiktextractContext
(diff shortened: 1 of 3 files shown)


daxida:
Note:

This template is not parsed properly (forms are not parsed at all). This happened before the changes in this PR.

https://ja.wiktionary.org/wiki/%E5%9C%A7%E3%81%99%E3%82%8B
https://kaikki.org/jawiktionary/All%20languages%20combined/meaning/%E5%9C%A7/%E5%9C%A7%E3%81%99/%E5%9C%A7%E3%81%99%E3%82%8B.html

```
===={{conjug}}====
{{日本語変格活用|圧|する|jsort=あっ}}
```

I fixed a similar one here:
https://ja.wiktionary.org/w/index.php?title=%E5%8F%B7%E3%81%99%E3%82%8B&action=history

but basically that template exposes the table (there is no dropdown), and the logic goes boom.

This is relevant because the category has the most items (but there are **other** templates in that category, so take it with a grain of salt)

https://ja.wiktionary.org/wiki/%E3%82%AB%E3%83%86%E3%82%B4%E3%83%AA:%E6%97%A5%E6%9C%AC%E8%AA%9E_%E5%8B%95%E8%A9%9E
image


This PR is finished, except for whatever feedback you may have about:

> table_name became unused (because it's always table1/table2, with no relevant information), and could be removed entirely, but I'd like to know what you think first.

xxyzz:
I think the first table shouldn't be discarded since it has different words than the second table.

Thanks for your contribution!

PR #1555, opened 2025-12-10
[de] Separate article in German pronouns

daxida:
Continuing from https://github.com/tatuylonen/wiktextract/pull/1549


xxyzz:
Thanks!

PR #1578, opened 2026-01-01
[en] Fix Japanese verb conjugation headers

0xcafeb33f:
Many headers for Japanese verb conjugation were missing, for example, なさる.

(please double-check the patch as I currently don't have access to a computer where I can test locally)


xxyzz:
Thanks!

PR #1608, opened 2026-03-11
fix: handle multi-word translations in German extractor

johnny:
Multi-word translations like {{Ü|fr|temps}} de {{Ü|fr|travail}} were being split into two separate entries ('temps' and 'travail') instead of being combined into 'temps de travail'.

Two changes:
- Plain text between Ü templates (e.g. 'de') is now concatenated to the current word instead of being silently dropped.
- Consecutive Ü templates with the same lang_code append their words to form a single multi-word translation, while different lang_codes still trigger a new translation entry.


kristian-clausal:
Apologies for merging this so late, I completely missed this! This seems fine to me, merged.

PR #1621, opened 2026-03-29
Use pytest

Darkdadaah:
I've run the default tests via unittest, and the output is quite messy (even though the tests run successfully).

I suggest switching to using pytest to run the tests, which doesn't necessitate any code change apart from the make file and pyproject.toml. There are several advantages, but the main one is that the Standard output/error are captured by pytest.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,10 +4,10 @@ REPO ?= tatuylonen/wiktextract
 SHA ?= HEAD
 
 test:
-	python -m unittest discover -b -s tests
+	python -m pytest tests
 test_coverage:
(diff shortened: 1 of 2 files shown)


xxyzz:
Test outputs are already captured by "unittest -b" option. The only output is from the "test_long.py" file, somehow the logger output in wikitextprocessor package is not captured. IMO a proper fix is to figure out what's wrong with the logger instead of add another dependency.

Darkdadaah:
Fair, I'll look into silencing that one properly

PR #1622, opened 2026-03-30
Clear image link regex cache for test_clean

Darkdadaah:
Because test_clean.py usually runs before the other language specific tests (by alphabetical chance) it implicitly does not reset its image link regex.
Switch the order of the tests, and test_clean fails.

This fix clears the cache to ensure the tests run from a clean slate.


kristian-clausal:
Thanks for catching this!

PR #1641, opened 2026-05-02
[en] Improve POS tag parsing in the pronunciation section

miyuki:
This series of two patches improves parsing of POS tags in the pronunciation section:
* The first patch makes the "pos" field a list rather than a str to support multiple POS tags attached to a single SoundData object and updates the parsing accordingly. It also adds support for line-local POS tags.
* The second patch adds support for POS tags in qualifier templates (such as `q` and `a`) in various scenarios (before and after pronunciations templates and as a tag for following lines).

### DIFF
--- a/src/wiktextract/extractor/en/page.py
+++ b/src/wiktextract/extractor/en/page.py
@@ -1021,6 +1021,16 @@ def parse_language(
                 pron.pop(key)  # type: ignore
             return pron
 
+        def sound_matches_pos(sound: SoundData, pos: str) -> bool:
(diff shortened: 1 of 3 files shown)


kristian-clausal:
I'm currently working on something in En pron (duplicated entries due to hidden sections), so I will finish that first and then take a look at this later. Thanks for your contribution!

kristian-clausal:
This looks great, it took me a while because I had to figure out the old code too. Merging this with another PR because I can't figure out how to do this properly.

PR #1648, opened 2026-05-13
[en] Improve POS tag parsing in the pronunciation section - part 2

miyuki:
The first commit fixes a small bug with over-eager replacement of "[...]" occurrences in qualifiers.

This the second commit addresses all remaining cases that I have discovered so far with POS markers in the pronunciation section:
* Unindexed (q=, qq=) and indexed (q1=, q2=) qualifier parameters in IPA and enPR templates
* Inline tags (<q:...>) in IPA and enPR templates

### DIFF
--- a/src/wiktextract/extractor/en/pronunciation.py
+++ b/src/wiktextract/extractor/en/pronunciation.py
@@ -62,6 +62,22 @@ class PronunciationPosPrefix(NamedTuple):
     is_persistent: bool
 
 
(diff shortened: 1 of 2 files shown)


miyuki:
I found an issue in the second commit. I'll fix it and re-submit.

kristian-clausal:
It's fine to keep the old PR and make it a draft until it's fully baked, we don't mind. Thanks for your (future) contribution. :)

PR #1649, opened 2026-05-14
[ja] Use `wago` as pos instead of `unknown`

daxida:
Title.

The rationale behind is that:
- "unknown" is lossy, and makes selecting this data potentionally risky when it conflicts with other headers that may also converge to "unknown". One could narrow the selection by only taking pos="unknown" and tags including "kanji" and "form-of", but that, also, is potentially faulty.
- "wago" is the name of the template that expands to this header. See [here](https://ja.wiktionary.org/wiki/%E9%8B%AD%E3%81%84). Unfortunately the template [page](https://ja.wiktionary.org/wiki/%E3%83%86%E3%83%B3%E3%83%97%E3%83%AC%E3%83%BC%E3%83%88:wago) is most succint. See also the wikipedia on [wago](https://en.wikipedia.org/wiki/Wago). See also the changed test below.

The obvious disadvantage is that "wago" doesn't really correspond to a "pos", but it is rather a convenient category for wiktionary editors. I don't think this is an issue since it practically works as a "pos", and that is how it is habitually used in these sort of redirection pages (which are many).

We could have opted for `wagokanji`, which wiktionary uses for the list items and the headwords, but strictly speaking, "wago" is the header, and, as shown above, it is also an accepted English word (albeit very technical).

We could have also removed the tags ("kanji" and "form-of"), but I wanted the changes to be minimal, even if they are (at least the "kanji" one), sort of redundant.

### DIFF
--- a/src/wiktextract/extractor/ja/section_titles.py
+++ b/src/wiktextract/extractor/ja/section_titles.py
@@ -31,7 +31,7 @@ POS_DATA = {
     # "小詞"
     # "修飾詞"
     # "疑問詞"
-    "和語の漢字表記": {"pos": "unknown", "tags": ["kanji", "form-of"]},
+    "和語の漢字表記": {"pos": "wago", "tags": ["kanji", "form-of"]},
     "成句": {"pos": "phrase", "tags": ["idiomatic"]},
     "成語": {"pos": "phrase", "tags": ["idiomatic"]},
     "縮約形": {"pos": "contraction"},
--- a/src/wiktextract/extractor/ja/section_titles.py
+++ b/src/wiktextract/extractor/ja/section_titles.py
@@ -31,7 +31,7 @@ POS_DATA = {
     # "小詞"
     # "修飾詞"
     # "疑問詞"
-    "和語の漢字表記": {"pos": "wago", "tags": ["kanji", "form-of"]},
+    "和語の漢字表記": {"pos": "character", "tags": ["wago-kanji", "form-of"]},
     "成句": {"pos": "phrase", "tags": ["idiomatic"]},
     "成語": {"pos": "phrase", "tags": ["idiomatic"]},
     "縮約形": {"pos": "contraction"},
--- a/src/wiktextract/tags.py
+++ b/src/wiktextract/tags.py
@@ -6133,6 +6133,7 @@ valid_tags = {
     "volitional": "mood",  # Verb mood (e.g., Japanese: suggests, urges, initates act)
     "vos-form": "register",  # Spanish verb forms used with "vos"
     "vulgar": "register",
+    "wago-kanji": "misc",
     "weak": "misc",
     "with-a": "with",
     "with-ablative": "with",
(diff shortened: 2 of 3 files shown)


kristian-clausal:
We have the PoS `character` used for kanji already, which fits here perfectly. We can easily create a new tag `wago-kanji` for word entries with vague PoSes representing kanji representations of wago words. Add `"wago-kanji": "misc",` in `valid_tags` in `tags.py` and should be good to go.

daxida:
Mm `character` is supposed to be for single character pages, no? That is not the case with `wago` pages. It's also debatable how relevant the "kanji" tag is (I assume xxyyz added it because the name of the `wagokanji` list template), since most of them are not **only** kanji. See the example above or [生き仏](https://ja.wiktionary.org/wiki/%E7%94%9F%E3%81%8D%E4%BB%8F) where the hiragana (not kanji) き appears on the word.

Also I changed the pos to `wago`, I'm not sure what adding `wago-kanji` to tags will do.

kristian-clausal:
`wago-kanji` is just more specific. `wago` could refer to kana-only representations of Japanese-native words, while `wago-kanji` would mean entries that use kanji at all.

For example, https://ja.wiktionary.org/wiki/%E3%81%84%E3%81%8D%E3%81%BC%E3%81%A8%E3%81%91 is the hiragana-representation of https://ja.wiktionary.org/wiki/%E7%94%9F%E3%81%8D%E4%BB%8F

It's just more specific, and using `wago-kanji` instead of the less specific `wago` does not cost anything except a little bit of memory and disk-space.

The definition of PoSes are not set in stone, and `character` seems '*fine*' to be used here. We do not want to have too-too many PoS categories. Wago-kanji entries like these are literally about kanji readings and nothing else (really, the kana are incidental for the purpose of the entry, well... almost), that's `character` material.

daxida:
So in short, you want:
- the pos to be `character`
- `wago-kanji` to be added to the tags of the `和語の漢字表記` section header
- And this, which I'm not entirely sure what it does:

> Add "wago-kanji": "misc", in valid_tags in tags.py

(which `tags.py` by the way?)

kristian-clausal:
Adding that line to "valid_tags" just registers the tag for certain purposes.

daxida:
Done. Not sure if I should have kept the `kanji` tag. I replaced it but let me know if you want me to add it back.

> Adding that line to "valid_tags" just registers the tag for certain purposes.

You know better than me. I thought that file was only for the English extractor and this is a Japanese only change but oh well.

daxida:
I just saw that a formating change slipped (blame autoformat). Let me know if that's an issue too.

kristian-clausal:
We want to keep all the extractors at least a bit in line with each other, you can consider `tags.py` to contain a lot of "should have" and "should be like this" stuff. Tags in `valid_tags` are valid tags for all extractors, so if we add a new tag like this for one extractor, we kind of have to do so for all of them.

I think leaving the "kanji" out could go either way, so it should be fine the way you've done it here. Thanks for your work with this!

kristian-clausal:
As long as the github linter doesn't complain, it's all fine.

PR #1654, opened 2026-05-25
Pronunciation POS tag parsing - part 2

miyuki:
This is the second part of my patch series related to improving part of speech parsing in the pronunciation section. It adds detection of POS markers in IPA and enPR templates in various contexts (indexed and unindexed a=/q= parameters, inline qualifiers, e.g. <q:...>).
It also handles the inheritance of POS tags in nested lists, e.g.
The first commit ([en] Adjust detection of stray /.../ and [...]) fixes the issue when some tags get replaced with "stripped-by-parse_pron_post_template_fn"

### DIFF
--- a/src/wiktextract/extractor/en/pronunciation.py
+++ b/src/wiktextract/extractor/en/pronunciation.py
@@ -27,6 +27,8 @@ from .form_descriptions import (
 from .parts_of_speech import part_of_speech_map
 from .type_utils import Hyphenation, SoundData, TemplateArgs, WordData
 
+PronunciationPoses = tuple[str, ...]
+
 # Prefixes, tags, and regexp for finding romanizations from the pronuncation
 # section
 pron_romanizations = {
@@ -52,14 +54,19 @@ IPA_EXTRACT_RE = re.compile(IPA_EXTRACT)
 
 
 class PronunciationPosMatch(NamedTuple):
-    pos_values: list[str]
+    pos_values: PronunciationPoses
     residual: str
 
 
 class PronunciationPosPrefix(NamedTuple):
-    pos_values: list[str]
+    pos_values: PronunciationPoses
     text: str
     is_persistent: bool
+
+
+class FlattenedListNode(NamedTuple):
+    node: WikiNode | str
+    list_depth: int
 
 
 PRON_POS_TEMPLATE_NAMES = {
@@ -74,6 +81,19 @@ PRON_POS_TEMPLATE_NAMES = {
     "lbl",
     "label",
 }
+
+PRON_POS_BY_LABEL = {
+    label: pos_data["pos"] for label, pos_data in part_of_speech_map.items()
+}
+
+PRON_POS_LABEL_RE = re.compile(
+    r"^(?:(?P<residual>.+?)\s+)?(?P<label>"
+    + "|".join(
+        re.escape(label)
+        for label in sorted(PRON_POS_BY_LABEL, key=len, reverse=True)
+    )
+    + r")$"
+)
 
 
 def normalize_pronunciation_pos_label(label: str) -> str:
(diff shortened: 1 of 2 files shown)


src/wiktextract/extractor/en/pronunciation.py
```diff
@@ -87,27 +99,92 @@ def normalize_pronunciation_pos_label(label: str) -> str:
     return label
 
 
+def pronunciation_pos_from_part(part: str) -> tuple[str | None, str]:
+    normalized = normalize_pronunciation_pos_label(part)
+    if normalized in PRON_POS_BY_LABEL:
+        return PRON_POS_BY_LABEL[normalized], ""
+    words = normalized.split()
+    first_label_word = max(1, len(words) - PRON_POS_MAX_LABEL_WORDS)
+    for i in range(first_label_word, len(words)):
```
kristian-clausal: It took me a while to understand what was happening in this loop. This would be probably clearer and faster as a regex. The `re.Pattern` can be generated instead of PRON_POS_MAX_LABEL_WORDS, by taking the keys from the dict and sorting them by descending length and `r"|".join()` ing them together as a long regex pattern, then extract `label` and `residual` with groups. Please also add a bit more documentation about what is going on.

src/wiktextract/extractor/en/pronunciation.py
```diff
@@ -87,27 +99,92 @@ def normalize_pronunciation_pos_label(label: str) -> str:
...
+            residual = " ".join(words[:i]).rstrip(" ,;:")
+            if residual and classify_desc(residual) != "tags":
+                continue
+            return PRON_POS_BY_LABEL[label], residual
+    return None, part
+
+
+def split_pronunciation_pos_parts(text: str) -> list[str]:
+    parts: list[str] = []
+    for comma_part in re.split(r"\s*[,;]\s*", text):
```
kristian-clausal: Because you're `.strip()` ing on line 121, the `\s*` are not necessary. Extremely minor optimization.

src/wiktextract/extractor/en/pronunciation.py
```diff
@@ -87,27 +99,92 @@ def normalize_pronunciation_pos_label(label: str) -> str:
...
+        comma_part = comma_part.strip()
+        if not comma_part:
+            continue
+        conjunction_parts = re.split(r"\s+\b(?:and|or)\b\s+", comma_part)
```
kristian-clausal: `\b` here doesn't do anything, it's always true when a space is followed by a letter or a letter is followed by a space.

src/wiktextract/extractor/en/pronunciation.py
```diff
@@ -87,27 +99,92 @@ def normalize_pronunciation_pos_label(label: str) -> str:
     return label
 
 
+def pronunciation_pos_from_part(part: str) -> tuple[str | None, str]:
+    normalized = normalize_pronunciation_pos_label(part)
```
kristian-clausal, line 132: Please add an example or two of input and output here to show what this function actually is targeting. EDIT: Found the comment later in the code, these functions could be moved after the function that calls them. We have a toooon of counter-examples of this order, but that's because Tatu comes from a heavy C-background.

src/wiktextract/extractor/en/pronunciation.py
```diff
@@ -87,27 +99,92 @@ def normalize_pronunciation_pos_label(label: str) -> str:
...
 
-def set_sound_pos(sound: SoundData, pos_values: list[str] | None) -> None:
-    if not pos_values:
-        return
-    sound["pos"] = list(pos_values)  # type: ignore[typeddict-unknown-key]
+def set_sound_pos(
+    sound: SoundData, pos_values: list[str] | None
+) -> list[str] | None:
+    if pos_values:
+        sound["pos"] = list(pos_values)  # type: ignore[typeddict-unknown-key]
```
kristian-clausal: I know this was like this in the original, but if we trust that `pos_values` is always `list[str]` the cast to list is unnecessary. Probably left over from a more general typing class like Sequence. We've shifted away from using typing generics because they've turned out to be overkill in most cases, and using concrete types like tuple or list is just simpler and easier. We don't expose any of our stuff outwards so the exact details of types don't matter.

src/wiktextract/extractor/en/pronunciation.py
```diff
@@ -621,9 +726,8 @@ def split_cleaned_node_on_newlines(
             text = re.sub(r"^\**\s*", "", text).strip()
             if i == 0:
                 # At the start of a line, check for stuff like "Noun:"
-                # for active_pos; active_pos is a temporary data field
-                # given to each saved SoundData entry which is later
```
kristian-clausal, line 624: is this comment moved or completely deleted?

src/wiktextract/extractor/en/pronunciation.py
```diff
@@ -872,13 +999,22 @@ def split_cleaned_node_on_newlines(
...
         # and those used to be stored under "hyphenation"
 
         # Add data that was collected in template_fn
+        # POS inheritance for audio has one extra source:
+        # common_sound_pos(line_sound_pos_candidates), from pronunciations
+        # extracted earlier on the same physical line, e.g.
+        # "* {{IPA|en|/foo/|a=verb}} {{audio|en|foo.wav}}".
+        # Explicit line_pos still wins, then same-line sound agreement, then
+        # parent-list structure, then active_pos.
+        audio_inherited_pos = (
```
kristian-clausal, line 1031: 👍

src/wiktextract/extractor/en/pronunciation.py
```diff
@@ -87,27 +99,92 @@ def normalize_pronunciation_pos_label(label: str) -> str:
...
+) -> list[str] | None:
+    if pos_values:
+        sound["pos"] = list(pos_values)  # type: ignore[typeddict-unknown-key]
+        return sound["pos"]  # type: ignore[typeddict-item]
+    if "pos" in sound:
+        return sound["pos"]  # type: ignore[typeddict-item]
+    return None
+
+
+def common_sound_pos(pos_candidates: list[list[str]]) -> list[str] | None:
```
kristian-clausal: This could be simplified a bit if the functions that return `list[str]` containing PoS variants instead returned `tuple[str]`; `set_sound_pos`, contents of `pos_pron_markers` <- `PronunciationPosMatch` <- `extract_pronunciation_pos_template()`, and everything downstream from that; I don't see any manipulation of these lists after they're finalized in `extract_pronunciation_pos_template()`.

With tuples, you can do a pretty efficient cast to a set with `unique_pos = list({pos_tuple for pos_tuple in pos_candidates})` or something similar... But mainly I just don't trust lists to behave anymore, tuples are much more nice to throw around as data.

kristian-clausal (changes requested):
Looking good, just left a few comments around code that took me a while to understand, or which could be made more idiomatic with our pre-existing code.

src/wiktextract/extractor/en/pronunciation.py
miyuki: The purpose of `list()` here is to create a copy so that we don't accidentally mutate unrelated instances of SoundData when we change pos in-place. I will fix this by using a tuple instead of list as you suggested in another comment.

kristian-clausal: I didn't think of that, even though I saw that exact same trick (list a list to get a new list) elsewhere, d'oh.

kristian-clausal: Because this is a kludge where we temporarily throw the pos list into the typeddict and remove it later, it probably doesn't break anything (I can't imagine we manipulate this list after this); it's just used to assign the sound data to its appropriate PoS contexts later, which should just be a for loop or an `in` check. If your typechecker complains, you can leave it for later and I'll take a look at it.

kristian-clausal:
I'm leaving the office right now, so I'll dip out of the conversation; I'll be available tomorrow again, but don't worry about that and take all the time you need with this.

src/wiktextract/extractor/en/pronunciation.py
```diff
@@ -27,6 +27,8 @@
 from .parts_of_speech import part_of_speech_map
 from .type_utils import Hyphenation, SoundData, TemplateArgs, WordData
 
+PronunciationPos = tuple[str, ...]
```
kristian-clausal: Minor thing, if you're still working on further commits, for clarity our project tries to use plural names for lists and tuples; so `PronunciationPoses` or `PronPOSes` or something similar. Not terribly important.

src/wiktextract/extractor/en/pronunciation.py
```diff
@@ -75,6 +82,19 @@ class PronunciationPosPrefix(NamedTuple):
     "label",
 }
 
+PRON_POS_BY_LABEL = {
+    label: pos_data["pos"] for label, pos_data in part_of_speech_map.items()
+}
+
+PRON_POS_LABEL_RE = re.compile(
+    r"^(?:(?P<residual>.+?)\s+)?(?P<label>"
```
kristian-clausal, line 90: I didn't know about Python regex group labels, that's neat!

kristian-clausal (approved):

kristian-clausal:
Looks good, can be merged, just say the word.

miyuki:
I've changed it to PronunciationPoses. Should be ready to merge.

miyuki:
Could you please merge it? I don't think I have permission to merge in this repo, even after approval.

kristian-clausal:
Sorry, missed your message! Thanks for your contribution!

PR #1661, opened 2026-05-30
[ja] Minor tag improvement

daxida:
Title.

To put it in line with other editions.

image


daxida:
Also note that this tag:
```json
    "首都": "uppercase",
```
is wrong but the fix is not easy.

It means capital as in capital city (for example: the page for [Reykjavík](https://ja.wiktionary.org/wiki/%E3%83%AC%E3%82%A4%E3%82%AD%E3%83%A3%E3%83%B4%E3%82%A3%E3%83%BC%E3%82%AF)), but the only capital registered in the English edition (and I assume that's how we ended with uppercase) does that conversion:

image

---

Maybe adding a `capital-city` tag to the English edition makes sense?

daxida:
Force pushed legal to law since it is done in English too, and the Japanese tag itself it's literally "law" and not "legal" anyway.

https://ja.wiktionary.org/wiki/%E6%B3%95%E5%BE%8B#Japanese

Used here:
https://ja.wiktionary.org/wiki/%E8%A1%8C%E6%B3%95

Also why are those conversions not global? That is, why is legal not converted to law in the Ja edition?

kristian-clausal:
If "首都" is 'capital city', it probably refers to Tokyo... Or, depending on context, some other (western) capital during an early period of Japanese history. I don't have context for it and don't know Japanese, can you figure out what it means specifically?

As for the "why's", it's because these were done by hand, and stuff will turn up that's wrong. If you find inconsistencies, please do bring them up.

daxida:
> If "首都" is 'capital city', it probably refers to Tokyo... Or, depending on context, some other (western) capital during an early period of Japanese history. I don't have context for it and don't know Japanese, can you figure out what it means specifically?

It seems to be used for modern cities. This is the category: https://ja.wiktionary.org/wiki/%E3%82%AB%E3%83%86%E3%82%B4%E3%83%AA:%E6%97%A5%E6%9C%AC%E8%AA%9E_%E9%A6%96%E9%83%BD%E5%90%8D

For example this page just stays: Dakar, capital of Senegal: https://ja.wiktionary.org/wiki/%E3%83%80%E3%82%AB%E3%83%BC%E3%83%AB

kristian-clausal:
So it's a (slightly weird) topical tag for senses that specify the entry is the name of a capital, not a linguistics tag. It should go into "topics", but we've been neglecting those.

EDIT: Committed and merged.

Issue #1630, opened 2026-04-16
[en] Mutation table has `error-unrecognized-form` tag


daxida:
Upstreaming this: https://github.com/yomidevs/wiktionary-to-yomitan/issues/342

The mutation table carries the `error-unrecognized-form`, even though the table seems parseable.

---

Some examples:

https://en.wiktionary.org/wiki/buneolas
https://kaikki.org/dictionary/Irish/meaning/b/bu/buneolas.html

https://en.wiktionary.org/wiki/adhmad#Irish
https://kaikki.org/dictionary/Irish/meaning/a/ad/adhmad.html

For instance, for this last one:

```json
    {
      "form": "adhmad",
      "source": "mutation",
      "tags": [
        "error-unrecognized-form" < should be "radical"
      ]
    },
    {
      "form": "n-adhmad",
      "source": "mutation",
      "tags": [
        "error-unrecognized-form" < should be "eclipsis"
      ]
    },
```

PR #1663, opened 2026-06-01
[en] whitelist some Czech inflection tags

### LINKED ISSUE #1655: [en] `error-unrecognized-form` in seemingly well-formed Czech tables
Title.

For example, in nebýt:
[wiktionary](https://en.wiktionary.org/wiki/nebýt)
[kaikki](https://kaikki.org/dictionary/All%20languages%20combined/meaning/n/ne/nebýt.html)

Image

---

For reference, upstreaming some details of this other issue: https://github.com/yomidevs/wiktionary-to-yomitan/issues/354

daxida:
Fixes #1655

I didn't know what to do with "active adjective/passive adjective" so I just copied the closest one (the one on top resp.). Feel free to change it if you think there's something more precise. They correspond to this:

image

---

and at any rate, it should be fine as long as we don't get the "error-unrecognized-tag".

This was done by updating some test in `test_en_inflection_cs` and eyeing the differences / reading the log messages.

Also this and #1630 are very similar in that wiktionary changes in the casing confuse the inflection parser. I browsed the map a bit, and I don't see any reason as to why the logic using this map couldn't convert to lower/upper (or whatever normalization you prefer) before searching the key. It does require tidying up infl_map but it would make these errors not happen in the future.


kristian-clausal:
I am pretty sure the capitalization distinction is kept because Tatu must have come across examples where that's meaningful. Technically, I guess we could do it by a "default" setup where the capitalization doesn't matter, unless there are two or more entries that differ, but that means you'd have to find all places in the code that check these things, and I'm not sure how easy it would be to do that correctly... Might be worth it.

EDIT: doing this by duplicating entries in the map isn't feasible, because we can only do it `text.lower()` transformations, because uppercase is unpredictable. You could have `Upper case` and `Upper Case` and `UPPER CASE`... Can't even every possible (and impossible) casing.

daxida:
A quick hack is to replace rename `infl_map` to `_infl_map` and add this after:

```py
class CaseInsensitiveDict(dict):
    def __getitem__(self, key):
        return super().__getitem__(key.lower())

    def __contains__(self, key):
        return super().__contains__(key.lower())

    def get(self, key, default=None):
        return super().get(key.lower(), default)


# at some point, delete _infl_map and only keep this
infl_map: CaseInsensitiveDict = CaseInsensitiveDict(
    {k.lower(): v for k, v in _infl_map.items()}
)
```

so that everything that touched infl_map can remain the same and you can see the result of running the tests to better assess if this is a worthwhile idea.

The LLM detected ambiguities:

```
"I": { ... },   # Roman numeral I
"i": { "lang": ["German", "Cimbrian"], "then": "subjunctive subjunctive-i", ... },
"II": { ... },
"ii": { "lang": ["German", "Cimbrian"], "then": "subjunctive subjunctive-ii", ... },
```

but from what I could gather from the tests, only positive diffs were shown (as in, missing tags that look that they should have been there to begin with, but I may be wrong). For instance:

image

---

Edit: There are also changes that seem like regressions. It's not entirely clear, at least to me.

kristian-clausal:
I'll talk with Tatu about this at some point.

Issue #1664, opened 2026-06-01
Consistent structure between extractors


daxida:
I'm currently doing some work that tries to revert the tag normalization to English (I know this is not going to end well already), and the first issue I've encountered is that the lack of consistency.

It is a matter of moving things around and renaming variables, that I would have to do anyway, but I figured it may be of use to have it directly here, so this issue is just to see if there is any interest in me committing changes in this direction.

For example, I need to know where the pos/tags information is located, so I made a script to scan for files and their variable names:

```
cs/section_titles.py:
  LINKAGE_SECTIONS, POS_DATA
cs/tags.py:
  GENDER_TAGS, LABEL_TAGS, SOUND_TAGS, TABLE_TAGS, TAGS, TOPICS

de/section_titles.py:
  FORM_TITLES
de/tags.py:
  CASE_TAGS, COMPARISON_TAGS, DECLENSION_TAGS, GENDER_TAGS, GRAMMATICAL_TAGS, INFLECTION_TABLE_TAGS, K_TEMPLATE_TAGS, K_TEMPLATE_TOPICS, MOOD_TAGS, NUMBER_TAGS, OTHER_TAGS, PERSON_TAGS, TENSE_TAGS, VERB_FORM_TAGS, VOICE_TAGS

el/section_titles.py:
  <NO DICT CONSTANTS>
el/tags.py:
  DONTCARES, ancient_greek_tables_tags, ancient_greek_tags, english_tables_tags, esperanto_tables_tags, french_tags, german_tables_tags, other_tags, polish_tables_tags, romanian_tables_tags, russian_tables_tags, spanish_tables_tags, transliteration_tags, turkish_tables_tags, zones_tags

es/section_titles.py:
  <NO DICT CONSTANTS>
es/tags.py:
  ALL_TAGS, AMBITO_TAGS, COMPARISON_TAGS, CSEM_TOPICS, ES_V_SUP_TAGS, GENDER_TAGS, NUMBER_TAGS, PERSON_TAGS, POS_TITLE_TAGS, SOUND_TAGS, TABLE_TAGS, TENSE_TAGS, USO_TAGS, VERB_FORM_TAGS

fr/section_titles.py:
  <FILE NOT FOUND>
fr/tags.py:
  ASPECT_TAGS, LEXIQUE_TAGS, ZH_PRON_TAGS

id/section_titles.py:
  LINKAGE_SECTIONS, POS_DATA
id/tags.py:
[...]

PR #1665, opened 2026-06-01
Rename to POS_DATA + add annotations if missing

daxida:
Deals with half of #1664

The reason for only addressing half of it is mainly to be easier to review. For that matter commits will be split into languages batches. The PR can be latter squashed if it becomes too noisy.

Namely this does only address the `POS_DATA` part. The convention, also stated in the [docs](https://github.com/tatuylonen/wiktextract/blob/master/docs/new_extractor_guide.md#create-section_titlespy-file) is to:
- have a `section_titles.py` file, with a `POS_DATA` python dictionary variable.
- (while not necessary) I added the typing of the dictionary because the ast traversal in my script differentiates between typed/non-typed dicts, and that I think it's a positive change overall.

This requires renaming variables, renaming files (in the case of `fr`), and adding the type if necessary.

---

Languages done (explained by commit):
- [x] cs, de, el, es, id
  - cs, es and id are annotations only
  - de, el are renames
- [x] it, ja, ko, ku, ms, nl, pt, sv, th, tr, vi (annotations)
- [x] zh, ru renames
  - ru is tricky since there is pos data in POS_TEMPLATE_NAMES too, but it makes more sense to rename POS_TITLES which was a common occurrence in other extractors.
- [x] `fr` rename (required a file rename too)
- [x] `simple` rename. This is the ONLY case I found where  we can't use `POSSubtitleData`. Therefore, the annotation didn't change.
- [x] `pl` required moving POS_DATA to a new section_titles.py (unexistent before)
  
---
  
Note that POS_TITLES is probably a better name for it (anything is better than data), but anyway, once they all share the same name, it becomes trivial to slam a global regex and change from POS_DATA to whatever else. It would require changing the docs too. 

Not only TITLES makes more sense, but prefixing instead of suffixing should be even better for autocompletion (TITLES_POS, TITLES_LINKAGE etc.). Again, not the goal of this PR.


kristian-clausal:
`el` and `simple` were mine, which is why the code-quality is lower than xxyzz's other editions, and the naming and typing are different. These two different streams were created independently, so there's differences. optional/missing fields for TypedDict are annoying, and getting the typing 100% right isn't necessary, because the extractors don't touch each other, so let's just leave that for a while until it becomes an issue, or the template data in `el` `POS_HEADINGS` could be moved into a parallel dict...

No wait, POSSubtitleData has `total=False`, we can just add `templates` there. If you see these kinds of overlapping types that are complementary and can be merged (if it makes sense), just point them out, they're not sacrosanct and can be changed. I can do it, or if you want you can do it, just say the word.

Looks good, thanks for the PR!

PR #1667, opened 2026-06-02
Rename to `TAGS` for consistency

daxida:
Continues the work on #1664

Following the [docs](https://github.com/tatuylonen/wiktextract/blob/master/docs/new_extractor_guide.md#create-tagspy-file), the convention is:
- a `tags.py` file with a `TAGS` dictionary
- I didn't check/add annotations this time for simplicity (and because the type is straightforward as opposed to POSSubtitleData, and values can vary between `str` and `str` | `list[str]`)

Note that `sv` doesn't have any tag data anywhere, so there's nothing to be done.

The convention asks for a `TOPICS` dictionary too, but that can wait for another PR.


kristian-clausal:
Looks good to me, thanks!

PR #1668, opened 2026-06-03
[ja] Improve `TAGS`

daxida:
Vaguely related to #1664

This is a side experiment where I replicated the testing logic with tags from the `el` edition for `ja` to see if it was practical.

It caught some subtle bugs, like the alternative hyphen in Wade-Giles, or anatomy being a topic and not a tag etc.

I don't know if adding a test for this is practical in every language, but if we added it for Japanese it will now pass (*except for the `capital-city` topic, see below).


daxida:
Also capital-city is not in the English data (not as a tag nor a topic), but since it was recently added I just ignored it.

kristian-clausal:
You should leave the overlapping `singular` and `plural` tags with the `singular-only`-style, more specific tags, that's the current style we have at the moment. There's no harm to the redundancy, and this case is so blatantly true that we can actually do it this way (which is not always true, but it's nice here).

daxida:
Well my reasoning is that I ripgrepped the repo and found that version to be more prevalent:

image

---

I don't mind changing it back though. I think the English edition is the one adding the raw plural sometimes.

image

kristian-clausal:
Thank you for the PR!

PR #1670, opened 2026-06-03
[ja] Resolve "呉音=*" in `ja-kanji` 

daxida:
It's documented here: https://ja.wiktionary.org/wiki/%E3%83%86%E3%83%B3%E3%83%97%E3%83%AC%E3%83%BC%E3%83%88:ja-kanji

> 呉音が漢音に共通する場合、呉音=*と記述できます。

that 呉音=* resolves to whatever value is in 漢音 (since it's a common occurrence).

This fixes that, to obtain the readings displayed here: https://ja.wiktionary.org/wiki/%E6%98%A5#Japanese

image

---

I added the test first, before the changes, for easier visualization (that test passed at that point, but ofc the result was not ideal).

### DIFF
--- a/src/wiktextract/extractor/ja/kanji.py
+++ b/src/wiktextract/extractor/ja/kanji.py
@@ -32,22 +32,44 @@ Kun'yomi (訓読み); native Japanese readings (hiragana):
 """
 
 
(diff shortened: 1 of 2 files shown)


kristian-clausal:
Good catch, thank you!

PR #1708, opened 2026-09-06
[pl,en] [fix] Ignored verb inflections

### LINKED ISSUE #1703: [pl-X] Polish tables missing inflection data
Expected: pobiegł to be in the json data of pobiec in both the en and the pl edition.

But there's nothing.

This should have been recent since it wasn't the case in July.

pobiec pobiegł

Polish edition
https://pl.wiktionary.org/wiki/pobiec
https://kaikki.org/plwiktionary/All%20languages%20combined/meaning/p/po/pobiec.html

English edition
https://en.wiktionary.org/wiki/pobiec
https://kaikki.org/dictionary/Polish/meaning/p/po/pobiec.html

See: https://github.com/yomidevs/wiktionary-to-yomitan/issues/379

daxida:
* They happen to not match a singular sense but the whole POS itself

Fixes #1703

Also not sure we want to add `sense_index = "1"`, since, again, it's matching the POS and so every sense inside it (1.1 to 1.4 for pobiec). Thoughts?

So the caveats are:
- [pl] Are we fine with `sense_index = "1"` even though there is none (senses go from 1.1 to 1.4; the "1" is the POS under which all of those sense fall into); or we would like to just parse the inflection and set `sense_index = ""`?
- [en] For the English fix, do we prefer storing the template names, or some heuristic down the line to determine the pre_expansion (both are flimsy, but storing the template name avoids a weird heuristic that can break any day)

### DIFF
--- a/src/wiktextract/extractor/pl/page.py
+++ b/src/wiktextract/extractor/pl/page.py
@@ -145,7 +145,7 @@ def match_sense_index(sense_index: str, word_entry: WordEntry) -> bool:
         if exact_match:
             return sense_index == sense.sense_index
 
-    pos_index_str = sense.sense_index[: sense_index.find(".")]
+    pos_index_str = sense.sense_index.split(".", 1)[0]
     pos_section_index = 0
     if pos_index_str.isdigit():
         pos_section_index = int(pos_index_str)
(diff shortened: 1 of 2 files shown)


kristian-clausal:
Sorry for this late comment, I missed this one and thought it was the other PR. This all seems fine. Just leave a comment here when you want to merge, so that I don't do it when you're still working on it.

daxida:
I was waiting for your feedback on the two points of the description. If you agree with both decisions then the work is done. If you want me to tweak something because you think the alternative (or some other approach) is more suitable, then I can revisit it.

kristian-clausal:
If I didn't mentioned anything last week, must have been fine (*trying to furiously remember what he said last week*)

PR #1712, opened 2026-09-18
[en] Fix parsing in japanese inflection tables

### LINKED ISSUE #1704: [ja-en] Regression in Japanese verb tables (English edition)
Many forms are now errors in verb tables.

```json
    {
      "form": "走らせる [hashiraseru]", // Expected: 走らせる
      "source": "conjugation",
      "tags": [
        "causative",
        "error-unrecognized-form" // Expected: this tag to not be present
      ]
    },
```

https://en.wiktionary.org/wiki/%E8%B5%B0%E3%82%8B
https://kaikki.org/dictionary/Japanese/meaning/%E8%B5%B0/%E8%B5%B0%E3%82%8B/%E8%B5%B0%E3%82%8B.html

See https://github.com/yomidevs/wiktionary-to-yomitan/commit/7b57612bb051c244bb26269f5d7b1e335fb515c9

daxida:
Fixes #1704

Pretty sure I'm missing some tags, but icba :D

Maybe there's a better solution by changing directly the parsing of the table.

This wasn't changed recently
https://en.wiktionary.org/w/index.php?title=Template:ja-conj-ex&action=history

but this was
https://en.wiktionary.org/w/index.php?title=Module:ja-conj&action=history

Let me know if you have better ideas for the solution. The test is what we want in any case, were we to accept this implementation or any other.


kristian-clausal (approved):
As always I cry,
when I see yet more kludges,
knowing they're needed.

-- Kurisu

